PR #1, opened 2015-08-13
Update coveralls integration.

nlutsenko:
Update the service for Coveralls, so it automatically is updated from Travis.


richardjrossiii:
LGTM

PR #5, opened 2015-08-13
Added automatic PFObject subclass registration.

richardjrossiii:
#### What's new ####

Currently, every `PFObject` subclass must be registered with parse before it is used. This can create some hard-to-track-down issues if you're not aware of this, and it isn't exactly the cleanest API.

In Objective-C/Swift, it is relatively easy to search at run-time for subclasses of `PFObject`, and have them register with parse automatically (in the case of dynamic libraries, even as soon as they load from their bundles).

This should prevent needing your app delegate to have lots of extra lines of code related to simply registering all of your `PFObject` subclasses.

#### What's breaking ####

 - If you had a subclass of `PFObject`, which implemented `PFSubclassing`, but intentionally did not register it with parse (or only conditionally registered it with parse), then you may run into errors printed to the console and a crash on start. The error message in the console should be concrete enough to help you track down the issue.
   - If you would still like to use the old-style manual registration (not recommended), you can opt-out on specific subclasses by conforming to the new protocol `PFSubclassingSkipAutomaticRegistration`. 

 - If you remove your calls to `[MySubclass registerSubclass]` in `AppDelegate` or similar, **and** `MySubclass` is located in a not-yet loaded dynamic library/framework, then until you manually load that library (by using a class/method/function inside it), your subclass will not be registered.
   - Note that this, however, allows for a much better potential plugin model using the Parse SDK using `dlopen` & friends, if your application needs such a feature.

These breaking changes shouldn't effect 99% of our users, and workarounds should be fairly simple for those it will break for.

### DIFF
--- a/Parse/Internal/Object/Subclassing/PFObjectSubclassingController.h
+++ b/Parse/Internal/Object/Subclassing/PFObjectSubclassingController.h
@@ -15,16 +15,10 @@
 @interface PFObjectSubclassingController : NSObject
 
 ///--------------------------------------
-#pragma mark - Init
-///--------------------------------------
-
-//TODO: (nlutsenko, richardross) Make it not terrible aka don't have singletons.
-+ (instancetype)defaultController;
-+ (void)clearDefaultController;
-
-///--------------------------------------
 #pragma mark - Registration
 ///--------------------------------------
+
+- (void)scanForUnregisteredSubclasses:(BOOL)shouldSubscribe;
 
 - (Class<PFSubclassing>)subclassForParseClassName:(NSString *)parseClassName;
 - (void)registerSubclass:(Class<PFSubclassing>)kls;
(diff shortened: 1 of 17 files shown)


richardjrossiii:
Fixed OSX unit tests, was an issue related to symlink resolution only happening partially.

bliang:
+1

richardjrossiii:
Rebased.

nlutsenko:
Hey guys, just letting everyone here know that we are tracking this to be merged in and release with version 1.8.4 of the SDK.

nlutsenko:
Looks like this needs conflict resolution.

vikmeup:
What's status on this? Thanks.

richardjrossiii:
@vikmeup We're waiting for version 1.11.0 until releasing it, due to wanting to get platform parity (implementing it on Java/.NET if possible) as well as perhaps breaking backwards compatibility.

nlutsenko:
Needs rebasing + tests to pass, back in your queue :)

iStopped:
+1

richardjrossiii:
Rebased.

richardjrossiii:
Due to issues with travis CI, we've moved this issue to #967, for any who are following this PR still.

PR #19, opened 2015-08-14
Use only frameworks from OCMock, convert to explicit reference.

nlutsenko:
`xcodebuild` is silly and non-deterministic when referencing 2 frameworks with the same name, which we currently have with OCMock.
Convert to using subproject references as well as start using dynamic framework for iOS target (not sure if better, just cleaner).


richardjrossiii:
LGTM... Sometimes, I just don't understand Xcode.

PR #20, opened 2015-08-14
Remove old GHUnit defines.

nlutsenko:
Convert the rest of defines to use XCTAssert, remove the old GHUnit-based ones as they are unused.


richardjrossiii:
LGTM.

PR #21, opened 2015-08-14
Re-add proper boolean support for object subclassing.

### LINKED ISSUE #13: BOOL <-> Number iOS Parse 1.8.0
Interesting issue: This was working ok on 1.7.5 before updating to 1.8.0
I have a subclass of PFUser with a BOOL property isActive and a NSNumber property hibernating
-- init as isActive = NO;
-- init hibernating = [NSNumber numberWithBool:NO];
once the first PFUser is created the isActive is a number not a BOOL and the hibernating NSNumber becomes a BOOL not a number!!!! 
Something is very wrong here, this was working correctly before.
Has anyone else seeing this problem?

richardjrossiii:
Fixes #13. This was broken due to `NSJSONSerialization` serializing all `NSNumbers` as '0' and '1', but any `CFBoolean`s as `true` and `false`. We now re-add explicit support for booleans.

### DIFF
--- a/Parse/Internal/Object/Subclassing/PFObjectSubclassingController.m
+++ b/Parse/Internal/Object/Subclassing/PFObjectSubclassingController.m
@@ -61,11 +61,28 @@ static CFNumberType PFNumberTypeForObjCType(const char *encodedType) {
         ['f'] = kCFNumberFloatType,
         ['d'] = kCFNumberDoubleType,
 
-        // C99 & CXX boolean
+        // C99 & CXX boolean. We are keeping this here for decoding, as you can safely use CFNumberGetBytes on a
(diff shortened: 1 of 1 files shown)


nlutsenko:
LGTM, single nit.

emonster:
Thank you for fixing this issue!!
Where do I get the fix? Can I get it by pod? or where can I download the new Parse.framework?

richardjrossiii:
You can change your pod file to use:

    pod 'Parse', :git = "https://github.com/parseplatform/parse-sdk-iOS-OSX.git"

Note that this will make you run off of master for parse, which is not guaranteed to be stable. You could specify a specific commit, details for which can be found on the cocoa pods website.

PR #29, opened 2015-08-17
Fix a crash when receiving a push notification with null sound

dominikkrejcik:
Previously there was a crash when receiving a push notification which had its sound set to `null`. This resulted in `soundName` being `NSNull` and crashed when checking for its length.


nlutsenko:
That is an amazing catch! Thank you so much!
Will merge in after Travis passes on it.

PR #31, opened 2015-08-17
Remove travis bundler caching to improve performance.

richardjrossiii:
Bundler caching was taking an obscene amount of time >2minutes, disable it.


nlutsenko:
LGTM. Thanks for making it faster!

PR #35, opened 2015-08-17
Rename PFObject '_state' instance variable to '_pfinternal_state'.

richardjrossiii:
**This is only a temporary workaround**. We should go back and clean up, not only our uses of the '_pfinternal_state' ivar directly, but also discuss whether our auto property synthesis still makes sense to have as a feature.

This prevents conflicts with our automatic property synthesis that is done for registered subclasses.

### DIFF
--- a/Parse/PFObject.m
+++ b/Parse/PFObject.m
@@ -103,7 +103,7 @@ static BOOL PFObjectValueIsKindOfMutableContainerClass(id object) {
     // Guards basically all of the variables below.
     NSObject *lock;
 
-    PFObjectState *_state;
+    PFObjectState *_pfinternal_state;
 
     PFObjectEstimatedData *_estimatedData;
     NSMutableSet *_availableKeys; // TODO: (nlutsenko) Maybe decouple this further.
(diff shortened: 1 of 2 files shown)


nlutsenko:
This works, but I am slightly skeptical on usage of double-underscores, since it's reserved by Apple.
It doesn't interfere with this, but definetly is something to be aware of.

@richardjrossiii Can we have a test for this, pretty please?

richardjrossiii:
Will write test, and change from double underscore. Stay tuned...

nlutsenko:
Still using double underscores for `__pfinternal`, do you think it's required?
Probably we can simply use `_pfinternal_state`, I don't think anyone is likely to use snake-case as well as prefix with `pfinternal` :D

richardjrossiii:
We already reserve fields in parse starting with `_`, so no, I guess the double underscore isn't necessary. I don't have any strong feelings either way, especially considering this is just a short-term workaround.

Will go with your lead and use `_pfinternal_state`.

nlutsenko:
LGTM, ship at will after nit is fixed.

PR #46, opened 2015-08-18
Store original network error under NSUnderlyingError key.

nlutsenko:
Still store in `originalError`, but add it to `NSUnderlyingErrorKey` as well.


richardjrossiii:
Errors, errors everywhere.

Issue #2, opened 2015-08-13
Combine podspecs, using platform-specific settings


segiddins:
i.e., in `Parse.podspec`, you could do

```ruby
s.ios.source_files = ...
s.osx.source_files = ...
```

That way, you wont need two separate specs.

nlutsenko:
Hey @segiddins, welcome... :D
Our OS X framework name is `ParseOSX`, and as Cocoapods doesn't support `s.osx.module_name` - we decided to leave it as is for this release, but will merge both of them together for the next one.

segiddins:
:+1: awesome -- you could just make `ParseOSX` depend on `Parse` and deprecate it

PR #48, opened 2015-08-18
Rename 'ParseOSX.framework' to 'Parse.framework' for OS X.

nlutsenko:
This will bring way better consistency to our code, as well as give us an ability to eventually merge podspecs together into a single one.

This is needed for #2.
Next stop - deprecate Parse-OSX podspec in favor of Parse.podspec


richardjrossiii:
veryseriouscodereview

PR #53, opened 2015-08-19
Deprecate Parse-OSX pod, add OS X support to Parse pod.

### LINKED ISSUE #2: Combine podspecs, using platform-specific settings

nlutsenko:
Closes #2 via next release.

### DIFF
--- a/Parse-OSX.podspec
+++ b/Parse-OSX.podspec
@@ -1,4 +1,7 @@
 Pod::Spec.new do |s|
+  s.deprecated = true
+  s.deprecated_in_favor_of = 'Parse'
+  
   s.name              = 'Parse-OSX'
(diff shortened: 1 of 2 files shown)


Parse-OSX.podspec
```diff
@@ -1,4 +1,7 @@
 Pod::Spec.new do |s|
+  s.deprecated = true
```
segiddins: this is redundant -- only use `deprecated_in_favor_of`

nlutsenko: TIL, thanks. Done in #55

Issue #52, opened 2015-08-19
Ability to revert all changes or a single field on `PFObject`.


nlutsenko:
After changing a single field on `PFObject` - there is no functionality to revert it back to the server state, even though we have the server state available internally. This is implemented in dotNet SDK via `ParseObject.revert()` method which reverts any changes to an object that were done after last successful save/fetch.

Implementation would require:
- Add `PFObject.revert()` that would clear all operations in `operationSetQueue` and rebuild `estimatedData`
- Add `PFObject.revertObjectForKey(key: String)` that would remove all operations for that key in `operationSetQueue` as well as rebuild `estimatedData`.

mykrr:
I am also facing the same issue as in #51. Please provide us quick fix for this.

kashif:
so something like:

```objective-c
- (void)revert {
    @synchronized (lock) {
        [operationSetQueue removeAllObjects];
        [self rebuildEstimatedData];
    }
}

-(void)revertObjectForKey:(NSString *)key {
    @synchronized (lock) {
        NSUInteger index = [operationSetQueue indexOfObject:key];
        [operationSetQueue removeObjectAtIndex:index];
        [self rebuildEstimatedData];
    }
}
```

nlutsenko:
@kashif pretty close, there are a little bit more things there - `availableKeys` would need to be adjusted as well as we need to `checkpointAllMutableContainers`.
Also - operationSetQueue contains PFOperationSet objects, not just keys.

kashif:
thanks for the info @nlutsenko just about to send a pull request. At the moment I am thinking:

```objective-c
- (void)revert {
    @synchronized (lock) {
        [operationSetQueue removeAllObjects];
        [self rebuildEstimatedData];
        [self checkpointAllMutableContainers];
    }
}

-(void)revertObjectForKey:(NSString *)key {
    @synchronized (lock) {
        NSUInteger index = [operationSetQueue indexOfObject:key];
        [operationSetQueue removeObjectAtIndex:index];
        [_availableKeys removeObject:key];
        
        PFFieldOperation *changes = [[self unsavedChanges] objectForKey:key];
        index = [operationSetQueue indexOfObject:changes];
        [operationSetQueue removeObjectAtIndex:index];
        
        [self rebuildEstimatedData];
        [self checkpointAllMutableContainers];
    }
}
```

grantland:
[...]

PR #70, opened 2015-08-20
Ability to revert all changes or changes for key

kashif:
For issue #52

### DIFF
--- a/Parse/PFObject.h
+++ b/Parse/PFObject.h
@@ -141,7 +141,7 @@ NS_REQUIRES_PROPERTY_DEFINITIONS
 /*!
  @abstract Sets the object associated with a given key.
 
- @param object The object for `key`. A strong reference to the object is maintaned by PFObject.
+ @param object The object for `key`. A strong reference to the object is maintained by PFObject.
  Raises an `NSInvalidArgumentException` if `object` is `nil`.
  If you need to represent a `nil` value - use `NSNull`.
  @param key The key for `object`.
@@ -176,7 +176,7 @@ NS_REQUIRES_PROPERTY_DEFINITIONS
  @discussion This method enables usage of literal syntax on `PFObject`.
  E.g. `object[@"key"] = @"value";`
 
- @param object The object for `key`. A strong reference to the object is maintaned by PFObject.
+ @param object The object for `key`. A strong reference to the object is maintained by PFObject.
  Raises an `NSInvalidArgumentException` if `object` is `nil`.
  If you need to represent a `nil` value - use `NSNull`.
  @param key The key for `object`.
@@ -201,6 +201,19 @@ NS_REQUIRES_PROPERTY_DEFINITIONS
  @deprecated Please use `[PFObject relationForKey:]` instead.
  */
 - (PFRelation *)relationforKey:(NSString *)key PARSE_DEPRECATED("Please use -relationForKey: instead.");
+
+/*!
+ @abstract Clears any changes to this object made since the last call to save and sets it back to the server state.
+ */
+- (void)revert;
+
+/*!
+ @abstract Clears any changes to this object's key that were done after last successful save and sets it back to the
+ server state.
+ 
+ @param key The key to revert changes for.
+ */
+- (void)revertObjectForKey:(NSString *)key;
 
 ///--------------------------------------
 /// @name Array Accessors
--- a/Parse/PFObject.m
+++ b/Parse/PFObject.m
@@ -2313,6 +2313,38 @@ static BOOL PFObjectValueIsKindOfMutableContainerClass(id object) {
     }
 }
 
+- (void)revert {
+    @synchronized (self.lock) {
+        if ([self isDirty]) {
+            NSMutableSet *persistentKeys = [NSMutableSet setWithArray:[self._state.serverData allKeys]];
+
+            PFOperationSet *unsavedChanges = [self unsavedChanges];
+            for (PFOperationSet *operationSet in operationSetQueue) {
+                if (operationSet != unsavedChanges) {
+                    [persistentKeys addObjectsFromArray:[operationSet.keyEnumerator allObjects]];
+                }
+            }
+
+            [unsavedChanges removeAllObjects];
+            [_availableKeys intersectSet:persistentKeys];
+
+            [self rebuildEstimatedData];
+            [self checkpointAllMutableContainers];
+        }
+    }
+}
+
+- (void)revertObjectForKey:(NSString *)key {
+    @synchronized (self.lock) {
+        if ([self isDirtyForKey:key]) {
+            [[self unsavedChanges] removeObjectForKey:key];
+            [self rebuildEstimatedData];
+            [_availableKeys removeObject:key];
+            [self checkpointAllMutableContainers];
+        }
+    }
+}
+
 #pragma mark Relations
 
 - (PFRelation *)relationforKey:(NSString *)key {
--- a/Tests/Unit/ObjectUnitTests.m
+++ b/Tests/Unit/ObjectUnitTests.m
@@ -10,6 +10,7 @@
 #import "PFObject.h"
 #import "PFUnitTestCase.h"
 #import "Parse_Private.h"
+#import "PFObjectPrivate.h"
 
 @interface ObjectUnitTests : PFUnitTestCase
 
@@ -186,7 +187,7 @@
 
 - (void)testFetchObjectWithoutObjectIdError {
     PFObject *object = [PFObject objectWithClassName:@"Test"];
-    
+
     XCTestExpectation *expectation = [self currentSelectorTestExpectation];
     [[object fetchInBackground] continueWithBlock:^id(BFTask *task) {
         XCTAssertNotNil(task.error);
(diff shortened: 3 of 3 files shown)


Parse/PFObject.h
```diff
@@ -813,6 +813,18 @@ NS_REQUIRES_PROPERTY_DEFINITIONS
  */
 - (BFTask *)deleteEventually;
 
+/*!
+ @abstract Reverts any changes to an object that were done after last successful save/fetch.
+ */
+ - (void)revert;
+
+/*!
+ @abstract Reverts any changes to an object's key that were done after last sucessful save/fetch.
```
grantland: This should actually be

> Clears any changes to this object made since the last call to save.

Since it won't revert anything prior to the current operation set in the queue.

Parse/PFObject.h
```diff
@@ -813,6 +813,18 @@ NS_REQUIRES_PROPERTY_DEFINITIONS
  */
 - (BFTask *)deleteEventually;
 
+/*!
```
nlutsenko: Can we move these to straight after `objectForKey:`, `setObjectForKey:`, as they are not really related to `**Eventually`.

Parse/PFObject.h
```diff
@@ -813,6 +813,18 @@ NS_REQUIRES_PROPERTY_DEFINITIONS
...
+ @abstract Clears any changes to this object made since the last call to save.
+ */
+ - (void)revert;
+
+/*!
+ @abstract Clears any changes to this object's key that were done after last successful save.
+ 
+ @param key The key to check for
```
nlutsenko: Nit: Missing `.` at the end.

Parse/PFObject.h
```diff
@@ -813,6 +813,18 @@ NS_REQUIRES_PROPERTY_DEFINITIONS
  */
 - (BFTask *)deleteEventually;
 
+/*!
+ @abstract Clears any changes to this object made since the last call to save.
+ */
+ - (void)revert;
+
+/*!
+ @abstract Clears any changes to this object's key that were done after last successful save.
```
nlutsenko: It's not `save` only, but rather back to the server state.

Parse/PFObject.m
```diff
@@ -2006,6 +2006,33 @@ - (void)deleteInBackgroundWithBlock:(PFBooleanResultBlock)block {
     [[self deleteInBackground] thenCallBackOnMainThreadWithBoolValueAsync:block];
 }
 
+- (void)revert {
+    @synchronized (lock) {
+        if ([self isDirty]) {
+            [operationSetQueue removeAllObjects];
```
nlutsenko: This should remove all the objects from the `unsavedChanges`, rather than the entire operationSetQueue, since the save operation could be in-flight right now and we don't want to revert that.

Parse/PFObject.m
```diff
@@ -2006,6 +2006,33 @@ - (void)deleteInBackgroundWithBlock:(PFBooleanResultBlock)block {
...
+            [self rebuildEstimatedData];
+            [self checkpointAllMutableContainers];
+        }
```
nlutsenko: This is also missing removal of keys that were reverted from `availableKeys`.

Parse/PFObject.m
```diff
@@ -2006,6 +2006,33 @@ - (void)deleteInBackgroundWithBlock:(PFBooleanResultBlock)block {
...
+
+-(void)revertObjectForKey:(NSString *)key {
+    @synchronized (lock) {
+        if ([self isDirtyForKey:key]) {
+            NSUInteger index = [operationSetQueue indexOfObject:key];
+            [operationSetQueue removeObjectAtIndex:index];
+            [_availableKeys removeObject:key];
+            
+            PFFieldOperation *changes = [[self unsavedChanges] objectForKey:key];
+            index = [operationSetQueue indexOfObject:changes];
```
nlutsenko: `[[self unsavedChanges] removeObjectForKey:key]`

nlutsenko:
@kashif Few nits and important parts, but overall looks amazing.

The important part here:
operationSetQueue is an NSMutableArray of PFOperationSet objects.
PFOperationSet is roughly close to NSMutableDictionary, which consists of `key`- > `change`.
What you are interested in is that last PFOperationSet, not modifying the operationSetQueue itself.

kashif:
thanks @nlutsenko gonna fix it up now

Parse/PFObject.h
```diff
@@ -201,6 +201,19 @@ NS_REQUIRES_PROPERTY_DEFINITIONS
...
+/*!
+ @abstract Clears any changes to this object made since the last call to save and sets it back to the server state.
+ */
+- (void)revert;
+
+/*!
+ @abstract Clears any changes to this object's key that were done after last successful save and sets it back to the
+ server state.
+ 
+ @param key The key to check for.
```
nlutsenko: Nit: `The key to revert changes for.`

Parse/PFObject.m
```diff
@@ -2317,6 +2317,36 @@ - (void)removeObjectForKey:(NSString *)key {
...
 
+- (void)revert {
+    @synchronized (lock) {
+        if ([self isDirty]) {
+            PFOperationSet *changes = [self unsavedChanges];
+            [changes enumerateKeysAndObjectsUsingBlock:^(id key, id obj, BOOL *stop) {
+                [_availableKeys removeObject:key];
+            }];
+            
+            [operationSetQueue removeLastObject];
```
nlutsenko: This can potential break a lot of things, since we rely on the fact that there is going to be a single `PFOperationSet` in the queue. `PFOperationSet` doesn't have `removeAllObjects` right now, but it would be easy to add.

Parse/PFObject.m
```diff
@@ -2317,6 +2317,36 @@ - (void)removeObjectForKey:(NSString *)key {
...
+            [self checkpointAllMutableContainers];
+        }
+    }
+}
+
+-(void)revertObjectForKey:(NSString *)key {
+    @synchronized (lock) {
+        if ([self isDirtyForKey:key]) {
+            NSUInteger index = [operationSetQueue indexOfObject:key];
+            [operationSetQueue removeObjectAtIndex:index];
```
nlutsenko: 2338-2339 - I don't think you need it at this point, since it won't find the index, as well as `removeObjectAtIndex:NSNotFound` (the index returned when it doesn't find anything) - will simply crash the app.

Parse/PFObject.m
```diff
@@ -2422,6 +2452,7 @@ - (NSString *)_flatDescription {
     }
 }
 
+
```
nlutsenko: Supernit: Remove this accidental change.

nlutsenko:
One more set of changes, almost there... Looks almost perfect.

@kashif 
Is there any chance you can also add some unit tests for these methods?
Tests around simple offline objects would be the minimal set, while it would be amazing to have tests using `_objectFromDictionary:defaultClassName:completeData:` - that one is used internally to create objects that are assumed to be fetched.

kashif:
Thanks @nlutsenko I will now try to add some unit tests

Parse/PFObject.m
```diff
@@ -2317,6 +2317,31 @@ - (void)removeObjectForKey:(NSString *)key {
     }
 }
 
+- (void)revert {
+    @synchronized (lock) {
+        if ([self isDirty]) {
+            PFOperationSet *changes = [self unsavedChanges];
+            for (NSString* key in changes.keyEnumerator) {
```
nlutsenko: Supernit: Use `for (NSString *key in...` to be closer to overall coding style.

Parse/PFObject.m
```diff
@@ -2317,6 +2317,31 @@ - (void)removeObjectForKey:(NSString *)key {
...
 
+- (void)revert {
+    @synchronized (lock) {
+        if ([self isDirty]) {
+            PFOperationSet *changes = [self unsavedChanges];
+            for (NSString *key in changes.keyEnumerator) {
+                [changes removeObjectForKey:key];
+            }
+            [self rebuildEstimatedData];
+            [_availableKeys removeAllObjects];
```
nlutsenko: It gets very tricky here, `_availableKeys` can contain keys that are in the serverData or in in-flight PFOperationSet (the one that is currently saving). I would suggest using all keys from estimated data + all keys from all operation sets (the reason for the latter, to avoid issue when the object was removed for any key and it would be gone from estimated data)

nlutsenko:
@kashif Restarted tests, and I think they are going to pass this time.
This looks very good, thanks for adding basic tests. There are few nits on the tests and one issue on the availableKeys, but we are close as ever to merging this in. I think a single iteration and we are done.

Please let me know if you have any questions around code or having trouble finding anything.

nlutsenko:
@kashif, sorry it takes a while to get to every single one of these, since there are literally 0 notifications on when the diff is pushed.
Let me know if I am wrong in my assumption or if you need/want any help.

kashif:
ok @nlutsenko i think it should pass the tests hopefully thanks again for your helpful suggestions!

nlutsenko:
Hey @kashif, so sorry, it looks like there is a merge conflict somewhere in code on this branch.
Can you please rebase against the latest master as well as squash all the commits into a single one?

kashif:
ok @nlutsenko rebased and squashed :+1:

nlutsenko:
@kashif Thank you so much for all the help, I am merging this one in and it's going to be released as part of 1.8.2

PR #81, opened 2015-08-21
Added weak/assign value support for PFPropertyInfo, and improved @synthesize detection.

richardjrossiii:
We now walk the class heirarchy for the synthesized variable. This may help in nested subclass use cases with a different synthesized name.


nlutsenko:
This needs documentation, but otherwise LGTM.

Issue #74, opened 2015-08-20
Status bar network no longer appears during Parse network activity


fatuhoku:
I've always seen it before the upgrade to 1.8.0. But now I don't see it.

This is desirable for debugging reasons as well as informing the user of network activity.

nlutsenko:
Hey @fatuhoku, thanks for reporting. Marking this for the next version, so we actually restore the network activity functionality.

PR #82, opened 2015-08-21
Added NSNotifications being posted on every network request being sent/finished.

nlutsenko:
This is required #74
- Added delegate to PFURLSession
- Updated Tests
- Added NSNotification posting whenever the request will be sent/finished.

[Review on Reviewable](https://reviewable.io/reviews/parseplatform/parse-sdk-ios-osx/82)

### DIFF
--- a/Parse/Internal/Commands/CommandRunner/URLSession/PFURLSessionCommandRunner.m
+++ b/Parse/Internal/Commands/CommandRunner/URLSession/PFURLSessionCommandRunner.m
@@ -30,6 +30,12 @@
 #import "PFURLConstructor.h"
 #import "PFURLSession.h"
 
+@interface PFURLSessionCommandRunner () <PFURLSessionDelegate>
+
+@property (nonatomic, strong) NSNotificationCenter *notificationCenter;
+
+@end
+
 @implementation PFURLSessionCommandRunner
 
 @synthesize applicationId = _applicationId;
@@ -49,11 +55,12 @@
                          clientKey:(NSString *)clientKey {
     NSURLSessionConfiguration *configuration = [[self class] _urlSessionConfigurationForApplicationId:applicationId
                                                                                             clientKey:clientKey];
-    PFURLSession *session = [PFURLSession sessionWithConfiguration:configuration];
+    PFURLSession *session = [PFURLSession sessionWithConfiguration:configuration delegate:self];
     PFCommandURLRequestConstructor *constructor = [PFCommandURLRequestConstructor constructorWithDataSource:dataSource];
     self = [self initWithDataSource:dataSource
                             session:session
-                 requestConstructor:constructor];
+                 requestConstructor:constructor
+                 notificationCenter:[NSNotificationCenter defaultCenter]];
     if (!self) return nil;
 
     _applicationId = [applicationId copy];
@@ -64,7 +71,8 @@
 
 - (instancetype)initWithDataSource:(id<PFInstallationIdentifierStoreProvider>)dataSource
                            session:(PFURLSession *)session
-                requestConstructor:(PFCommandURLRequestConstructor *)requestConstructor {
+                requestConstructor:(PFCommandURLRequestConstructor *)requestConstructor
+                notificationCenter:(NSNotificationCenter *)notificationCenter {
     self = [super init];
     if (!self) return nil;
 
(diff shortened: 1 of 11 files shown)


richardjrossiii:
I'm not super happy about using the default `NSNotificationCenter` everywhere. Thoughts on having a Parse-specific dependency-injected one? It would make testing/mocking better I think.

Parse/Internal/Commands/CommandRunner/URLSession/Session/PFURLSession.h
```diff
@@ -17,15 +17,29 @@
 
 NS_ASSUME_NONNULL_BEGIN
 
+@class PFURLSession;
+
+@protocol PFURLSessionDelegate <NSObject>
+
+- (void)urlSession:(PFURLSession *)session willPerformURLRequest:(NSURLRequest *)request;
+- (void)urlSession:(PFURLSession *)session didPerformURLRequest:(NSURLRequest *)request withURLResponse:(nullable NSURLResponse *)response;
```
richardjrossiii: If we fail below the protocol level (e.g. user switches wifi off), this will be called without a URL Response, and no way to get the error out. Thoughts on having a way to get the underlying network error if one exists here?

There's also no good way to track cancelled requests.

Parse/Internal/Commands/CommandRunner/URLSession/Session/PFURLSession.m
```diff
@@ -161,14 +167,18 @@ - (BFTask *)performFileDownloadURLRequestAsync:(NSURLRequest *)request
...
+    [self.delegate urlSession:self willPerformURLRequest:dataTask.originalRequest];
+
     @weakify(self);
     return [BFTask taskFromExecutor:[BFExecutor defaultExecutor] withBlock:^id{
         @strongify(self);
         NSNumber *taskIdentifier = @(dataTask.taskIdentifier);
         [self setDelegate:delegate forDataTask:dataTask];
 
-        BFTask *resultTask = [delegate.resultTask continueWithBlock:^id(BFTask *task) {
+        BFTask *resultTask = [delegate.resultTask continueWithExecutor:[BFExecutor defaultPriorityBackgroundExecutor] withBlock:^id(BFTask *task) {
```
richardjrossiii: Why did we change the executor here?

nlutsenko:
Parse-specific - not really
Dependency-injected - sure

Will update in a bit.

Parse/Internal/Commands/CommandRunner/URLSession/Session/PFURLSession.m
nlutsenko: Changed the executor - make sure the delegate callback is made on asynchronous thread and never inline (default executor can execute immediately in 19/20 chances).
`continueAsyncWithBlock` - good point.

Parse/Internal/Commands/CommandRunner/URLSession/Session/PFURLSession.h
nlutsenko: For this specific case - we don't really care about the result/error/cancellation, only about when the request is about to be sent and when request succeeded.
We don't really need an error right now, or a result, or cancelled status - that's why these are not implemented.

We can expand from here though at any given point of time.

Parse/Internal/Commands/CommandRunner/URLSession/PFURLSessionCommandRunner.m
```diff
@@ -49,11 +55,12 @@ - (instancetype)initWithDataSource:(id<PFInstallationIdentifierStoreProvider>)da
                          clientKey:(NSString *)clientKey {
     NSURLSessionConfiguration *configuration = [[self class] _urlSessionConfigurationForApplicationId:applicationId
                                                                                             clientKey:clientKey];
-    PFURLSession *session = [PFURLSession sessionWithConfiguration:configuration];
+    PFURLSession *session = [PFURLSession sessionWithConfiguration:configuration delegate:self];
```
richardjrossiii: This is bad. Self may be re-assigned by `init`, and then we have a dead delegate value. We should think about how to improve this.

nlutsenko: Sure, the possibility of that happening is really really close to 0, so better testability (with injected everything) vs rare edge case scenario.

PR #83, opened 2015-08-21
Add automatic management of Network Activity Indicator on iOS.

### LINKED ISSUE #74: Status bar network no longer appears during Parse network activity

nlutsenko:
Fixes #74 
Depends on #82 

Just subscribe it to the network indicator.

### DIFF
--- a/Parse/PFNetworkActivityIndicatorManager.m
+++ b/Parse/PFNetworkActivityIndicatorManager.m
@@ -10,6 +10,7 @@
 #import "PFNetworkActivityIndicatorManager.h"
 
 #import "PFApplication.h"
+#import "PFCommandRunningConstants.h"
 
 static NSTimeInterval const PFNetworkActivityIndicatorVisibilityDelay = 0.17;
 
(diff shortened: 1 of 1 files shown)


Parse/Internal/Commands/CommandRunner/URLSession/Session/PFURLSession.h
```diff
@@ -17,15 +17,29 @@
 
 NS_ASSUME_NONNULL_BEGIN
 
+@class PFURLSession;
+
+@protocol PFURLSessionDelegate <NSObject>
+
+- (void)urlSession:(PFURLSession *)session willPerformURLRequest:(NSURLRequest *)request;
+- (void)urlSession:(PFURLSession *)session didPerformURLRequest:(NSURLRequest *)request withURLResponse:(nullable NSURLResponse *)response;
```
richardjrossiii: If we fail below the protocol level (e.g. user switches wifi off), this will be called without a URL Response, and no way to get the error out. Thoughts on having a way to get the underlying network error if one exists here?

richardjrossiii:
Other than that the overall solution looks good. I'm not super keen on just using `NSNotificationCenter` all over the place, especially because it makes things harder to test/mock. Thoughts on having a Parse-specific `NSNotificationCenter` instance, that we can inject? Maybe its overkill.

PR #109, opened 2015-08-26
Make sure we compile for App Extensions from source.

### LINKED ISSUE #43: Not compiling when used with Extensions.
I keep getting sharedApplication is unavailable on iOS (App Extension).
This has happened in PFPush, PFApplication, PFPushUtilities, PFLocationManager and PFNetworkActivityIndicatorManager. I am also getting ld: warning: linking against dylib not safe for use in application extensions:

nlutsenko:
- Mark extension unavailable APIs
- Add workaround for UIApplication/NSApplication sharedApplication (it will be nil anyway)

Tested that it compiles for extensions by enabling `allow extension only APIs on both targets`.
We probably should have a sample application with a sample extension inside, so we can test this on a CI.
Fixes #43 though, since we fully compile for app extensions now.

### DIFF
--- a/Parse/Internal/PFApplication.h
+++ b/Parse/Internal/PFApplication.h
@@ -9,11 +9,20 @@
 
 #import <Foundation/Foundation.h>
 
+#if TARGET_OS_IPHONE
+#import <UIKit/UIKit.h>
+#else
+#import <AppKit/AppKit.h>
+@compatibility_alias UIApplication NSApplication;
+#endif
+
 /*!
  `PFApplication` class provides a centralized way to get the information about the current application,
  or the environment it's running in. Please note, that all device specific things - should go to <PFDevice>.
  */
 @interface PFApplication : NSObject
+
+@property (nonatomic, strong, readonly) UIApplication *systemApplication;
 
 @property (nonatomic, assign, readonly, getter=isAppStoreEnvironment) BOOL appStoreEnvironment;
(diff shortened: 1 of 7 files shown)


Parse/Internal/PFApplication.h
```diff
@@ -9,12 +9,21 @@
 
 #import <Foundation/Foundation.h>
 
+#if TARGET_OS_IPHONE
+#import <UIKit/UIKit.h>
+#else
+#import <AppKit/AppKit.h>
+@compatibility_alias UIApplication NSApplication;
```
richardjrossiii: Where's my holy water when I need it?

Parse/Internal/PFApplication.m
```diff
@@ -72,11 +72,20 @@ - (NSInteger)iconBadgeNumber {
...
+        self.systemApplication.applicationIconBadgeNumber = iconBadgeNumber;
 #else
         [[NSApplication sharedApplication] dockTile].badgeLabel = [@(iconBadgeNumber) stringValue];
 #endif
     }
 }
 
+- (UIApplication *)systemApplication {
+    // Workaround to make `sharedApplication` still be called even if compiling for App Extensions or WatchKit apps.
+#if TARGET_OS_IPHONE
```
richardjrossiii: Because of the compat alias, you shouldn't need this `#if`..

PR #120, opened 2015-08-27
Checks for nil applicationId and clientKey

gonzalonunez:
Searched over the TODOs and found this. The reason I passed the nil string 'yolo' in the tests is because if you actually pass in nil you'll get a warning.

I don't have too much professional experience with unit tests, please let me know if there's anything wrong with these tests.

### DIFF
--- a/Tests/Unit/ParseSetupUnitTests.m
+++ b/Tests/Unit/ParseSetupUnitTests.m
@@ -45,4 +45,12 @@
     [Parse setApplicationId:@"a" clientKey:@"b"];
     PFAssertThrowsInconsistencyException([Parse enableLocalDatastore]);
 }
+
+- (void)testInitializeWithNilApplicationIdNilClientKeyShouldThrowException {
+    NSString *yolo = nil;
+    PFAssertThrowsInconsistencyException([Parse setApplicationId:yolo clientKey:yolo]);
+    PFAssertThrowsInconsistencyException([Parse setApplicationId:yolo clientKey:@"a"]);
+    PFAssertThrowsInconsistencyException([Parse setApplicationId:@"a" clientKey:yolo]);
+}
+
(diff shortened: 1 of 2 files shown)


richardjrossiii:
Looks good to me! I'll let @nlutsenko take a peek and get his thoughts before merging.

Tests/Unit/ParseSetupUnitTests.m
```diff
@@ -45,4 +45,22 @@ - (void)testInitializeWithLDSAfterInitializeShouldThrowException {
...
 }
+
+- (void)testInitializeWithNilApplicationIdNilClientKeyShouldThrowException {
+    NSString *yolo = nil;
+    PFAssertThrowsInconsistencyException([Parse setApplicationId:yolo clientKey:yolo]);
+}
+
+- (void)testInitializeWithNilApplicationIdShouldThrowException {
+    NSString *yolo = nil;
+    PFAssertThrowsInconsistencyException([Parse setApplicationId:yolo clientKey:@"a"]);
```
richardjrossiii: If you want to avoid the temporary variable, you can cast `nil` to an `id` to suppress the warning IIRC.

nlutsenko:
Hey @gonzalonunez, thanks for implementing this!
Since all of these tests are single line assertions - can you please merge them into a single test with 3 assertions? It's testing roughly the same thing.

gonzalonunez:
Ok sounds good, I'm on it!

nlutsenko:
@gonzalonunez, last request - can you squash everything into a single commit? I will merge it straight after that is done. Thanks for the help with this one!

gonzalonunez:
Squashed 👍

nlutsenko:
Perfect! Merging it in!

PR #122, opened 2015-08-27
Cleaned up some warnings from the static analyzer

### LINKED ISSUE #121: Static Analyzer Warnings
I ran it through the static analyzer and some stuff came up. Thoughts?

screen shot 2015-08-27 at 4 48 22 pm

The first one is probably a compiler blip since remoteOperationSet is originally initialized as nil. The line in question is line 1116 of PFObject.m. 

The rest are just some coding convention issues with any methods that accept NSError** as a parameter and have a void return value. See:

screen shot 2015-08-27 at 4 53 11 pm

Is this anything worth fixing? I'd like to hear your guys' thoughts on this before going ahead and rewriting some stuff.

gonzalonunez:
Fixes #121

### DIFF
--- a/Parse/PFObject.h
+++ b/Parse/PFObject.h
@@ -491,7 +491,7 @@ NS_REQUIRES_PROPERTY_DEFINITIONS
 
  @deprecated Please use `-fetch` instead.
  */
-- (void)refresh PARSE_DEPRECATED("Please use `-fetch` instead.");
+- (instancetype)refresh PARSE_DEPRECATED("Please use `-fetch` instead.");
 
 /*!
  @abstract *Synchronously* refreshes the `PFObject` with the current data from the server and sets an error if it occurs.
@@ -500,7 +500,7 @@ NS_REQUIRES_PROPERTY_DEFINITIONS
 
  @deprecated Please use `-fetch:` instead.
  */
-- (void)refresh:(NSError **)error PARSE_DEPRECATED("Please use `-fetch:` instead.");
+- (instancetype)refresh:(NSError **)error PARSE_DEPRECATED("Please use `-fetch:` instead.");
 
 /*!
  @abstract *Asynchronously* refreshes the `PFObject` and executes the given callback block.
@@ -531,13 +531,13 @@ NS_REQUIRES_PROPERTY_DEFINITIONS
 /*!
  @abstract *Synchronously* fetches the PFObject with the current data from the server.
  */
-- (void)fetch;
+- (instancetype)fetch;
 /*!
  @abstract *Synchronously* fetches the PFObject with the current data from the server and sets an error if it occurs.
 
  @param error Pointer to an `NSError` that will be set if necessary.
  */
-- (void)fetch:(NSError **)error;
+- (instancetype)fetch:(NSError **)error;
 
 /*!
  @abstract *Synchronously* fetches the `PFObject` data from the server if <isDataAvailable> is `NO`.
@@ -614,7 +614,7 @@ NS_REQUIRES_PROPERTY_DEFINITIONS
 
  @param objects The list of objects to fetch.
  */
-+ (void)fetchAll:(PF_NULLABLE NSArray *)objects;
++ (NSArray *)fetchAll:(PF_NULLABLE NSArray *)objects;
 
 /*!
  @abstract *Synchronously* fetches all of the `PFObject` objects with the current data from the server
@@ -623,13 +623,13 @@ NS_REQUIRES_PROPERTY_DEFINITIONS
  @param objects The list of objects to fetch.
  @param error Pointer to an `NSError` that will be set if necessary.
  */
-+ (void)fetchAll:(PF_NULLABLE NSArray *)objects error:(NSError **)error;
++ (NSArray *)fetchAll:(PF_NULLABLE NSArray *)objects error:(NSError **)error;
 
 /*!
  @abstract *Synchronously* fetches all of the `PFObject` objects with the current data from the server.
  @param objects The list of objects to fetch.
  */
-+ (void)fetchAllIfNeeded:(PF_NULLABLE NSArray *)objects;
++ (NSArray *)fetchAllIfNeeded:(PF_NULLABLE NSArray *)objects;
 
 /*!
  @abstract *Synchronously* fetches all of the `PFObject` objects with the current data from the server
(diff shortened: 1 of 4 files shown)


Parse/PFObject.h
```diff
@@ -638,7 +638,7 @@ NS_REQUIRES_PROPERTY_DEFINITIONS
  @param objects The list of objects to fetch.
  @param error Pointer to an `NSError` that will be set if necessary.
  */
-+ (void)fetchAllIfNeeded:(PF_NULLABLE NSArray *)objects error:(NSError **)error;
++ (instancetype)fetchAllIfNeeded:(PF_NULLABLE NSArray *)objects error:(NSError **)error;
```
nlutsenko: This one should have an `NSArray` return type, since the block has an `NSArray` as a result.

Parse/PFObject.h
```diff
@@ -623,13 +623,13 @@ NS_REQUIRES_PROPERTY_DEFINITIONS
...
  */
-+ (void)fetchAll:(PF_NULLABLE NSArray *)objects error:(NSError **)error;
++ (instancetype)fetchAll:(PF_NULLABLE NSArray *)objects error:(NSError **)error;
 
 /*!
  @abstract *Synchronously* fetches all of the `PFObject` objects with the current data from the server.
  @param objects The list of objects to fetch.
  */
-+ (void)fetchAllIfNeeded:(PF_NULLABLE NSArray *)objects;
++ (instancetype)fetchAllIfNeeded:(PF_NULLABLE NSArray *)objects;
```
nlutsenko: This one should have an `NSArray` return type, since the block has an `NSArray` as a result.

Parse/PFObject.h
```diff
@@ -623,13 +623,13 @@ NS_REQUIRES_PROPERTY_DEFINITIONS
  @param objects The list of objects to fetch.
  @param error Pointer to an `NSError` that will be set if necessary.
  */
-+ (void)fetchAll:(PF_NULLABLE NSArray *)objects error:(NSError **)error;
++ (instancetype)fetchAll:(PF_NULLABLE NSArray *)objects error:(NSError **)error;
```
nlutsenko: This one should have an `NSArray` return type, since the block has an `NSArray` as a result.

Parse/PFObject.h
```diff
@@ -614,7 +614,7 @@ NS_REQUIRES_PROPERTY_DEFINITIONS
 
  @param objects The list of objects to fetch.
  */
-+ (void)fetchAll:(PF_NULLABLE NSArray *)objects;
++ (instancetype)fetchAll:(PF_NULLABLE NSArray *)objects;
```
nlutsenko: This one should have an `NSArray` return type, since the block has an `NSArray` as a result.

Parse/PFObject.m
```diff
@@ -1110,7 +1110,7 @@ - (void)mergeFromRESTDictionary:(NSDictionary *)object withDecoder:(PFDecoder *)
                 if (localOperationSet.updatedAt != nil &&
                     [localOperationSet.updatedAt compare:remoteOperationSet.updatedAt] != NSOrderedAscending) {
                     [localOperationSet mergeOperationSet:remoteOperationSet];
-                } else {
+                } else if (remoteOperationSet) {
```
nlutsenko: This one is actually very tricky, since `remoteOperationSet` should always be here, according to the logic above.
Maybe replace it with `PFConsistencyAssert` with a useful error message?
I just checked locally - it actually supresses the analyzer warning if you have an assert here.

nlutsenko:
@gonzalonunez, almost there...
Please fix inconsistency in return types between header and implementation files (that's what fails travis)

Parse/PFObject.m
gonzalonunez: Yeah, I wasn't too sure about that either. I'll go ahead and do that instead!

gonzalonunez:
Should've fixed it all. Waiting to make sure Travis doesn't fail.

gonzalonunez:
Seems to be missing something. Checking over it now.

nlutsenko:
Looks like it's related to `refresh` methods signature.

nlutsenko:
Ow, right... Those methods are declared also in `PFObjectPrivate.h` - that's what is missing.

gonzalonunez:
Ah, great catch. I'll go ahead and change those now. Thanks for the tip, would've taken me all night! Lol

Parse/PFObject.m
```diff
@@ -1111,6 +1111,7 @@ - (void)mergeFromRESTDictionary:(NSDictionary *)object withDecoder:(PFDecoder *)
                     [localOperationSet.updatedAt compare:remoteOperationSet.updatedAt] != NSOrderedAscending) {
                     [localOperationSet mergeOperationSet:remoteOperationSet];
                 } else {
+                    PFConsistencyAssert(remoteOperationSet, @"'remoteOperationSet' should not be nil.");
```
nlutsenko: <3

Parse/PFUser.m
```diff
@@ -1113,11 +1113,11 @@ - (BFTask *)fetchAsync:(BFTask *)toAwait {
     }];
 }
 
-- (void)fetch:(NSError **)error {
+- (instancetype)fetch:(NSError **)error {
     if (self.isLazy) {
-        return;
+        return nil;
```
nlutsenko: Still return `self` here, though...
Returning `nil` would mean that there was an error, but the idea here is basically that the object doesn't need to be fetched, since it's lazy.

Like in `fetchInBackgroundWithBlock:` we have
```objc
- (void)fetchInBackgroundWithBlock:(PFObjectResultBlock)block {
    if (self.isLazy) {
        if (block) {
            block(self, nil);
            return;
        }
    }
    [super fetchInBackgroundWithBlock:^(PFObject *result, NSError *error) {
        if (block) {
            block(result, error);
        }
    }];
}
```

nlutsenko:
@gonzalonunez, almost there, a single change around `fetch` method in PFUser and we are good to go.

Parse/PFUser.m
gonzalonunez: Ah, yes.Totally should've seen that, rookie move. On it!

gonzalonunez:
Fixed the issue in PFUser.m and squashed the commits 🙌

nlutsenko:
Looks perfect! Merging now!

PR #124, opened 2015-08-28
Implemented PFObject.deleteAll through PFObjectBatchController.

nlutsenko:
- Implemented PFObject.deleteAll through PFObjectBatchController
- Tested against api.parse.com - it works beautifully
- Way better decoupled state, as well as properly using dependency injected command runner
- Fixed bugs around `deleteAll` not properly deleting objects from LDS, as well as not updating the `deleted` property on state

### DIFF
--- a/Parse/Internal/Object/BatchController/PFObjectBatchController.m
+++ b/Parse/Internal/Object/BatchController/PFObjectBatchController.m
@@ -9,6 +9,8 @@
 
 #import "PFObjectBatchController.h"
 
+#import <Bolts/Bolts.h>
+
 #import "BFTask+Private.h"
 #import "PFAssert.h"
 #import "PFCommandResult.h"
@@ -19,6 +21,8 @@
 #import "PFObjectPrivate.h"
 #import "PFQueryPrivate.h"
 #import "PFRESTQueryCommand.h"
+#import "PFRESTObjectCommand.h"
+#import "PFRESTObjectBatchCommand.h"
 
 @implementation PFObjectBatchController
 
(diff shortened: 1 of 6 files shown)


nlutsenko:
This still needs some cleanup and tests, but the implementation is about as concrete as it would get.

nlutsenko:
All done, waiting for you review.

Parse/Internal/Object/BatchController/PFObjectBatchController.m
```diff
@@ -99,9 +103,84 @@ - (BFTask *)_processFetchResultAsync:(NSDictionary *)result forObjects:(NSArray
...
+}
+
+- (BFTask *)_processDeleteResultsAsync:(NSArray *)results forObjects:(NSArray *)objects {
+    NSMutableArray *tasks = [NSMutableArray arrayWithCapacity:results.count];
+    [results enumerateObjectsUsingBlock:^(NSDictionary *result, NSUInteger idx, BOOL *stop) {
+        PFObject *object = objects[idx];
+        NSDictionary *errorResult = result[@"error"];
+        NSDictionary *successResult = result[@"success"];
+
+        id<PFObjectControlling> controller = [[object class] objectController];
```
richardjrossiii: Don't we have access to the data source here? Why go through the class dispatch method.

nlutsenko: Objects can have different object controllers, say UserController or InstallationController, where those have extra logic on top of the bare object controller.

Tests/Unit/ObjectBatchControllerTests.m
```diff
@@ -172,4 +257,21 @@ - (void)testUniqueObjectsValidation {
                                                                              omitObjectsWithData:YES]));
 }
 
+- (void)testUniqueObjectsUsingFilter {
```
richardjrossiii: Let's test the duplicate use case as well.

nlutsenko: Already done.

PR #132, opened 2015-08-28
Use 'self' instead of direct class name in PFObject, PFUser.

nlutsenko:
Use `self` instead of using `PFObject` or `PFUser` class names.


richardjrossiii:
This is such a selfless diff.

Issue #133, opened 2015-08-28
Add nullability annotations for all private headers


nlutsenko:
We currently have 100% nullability annotations for all public headers, but there are quite a lot of private ones that are missing this.
This issue is to track addition of nullability annotations to all of them.
The plan for every one of those headers is as follows:
- Use `NS_ASSUME_NONNULL_BEGIN` and `NS_ASSUME_NONNULL_END` for entire header
- Add `nullable` to all types that can be `nil` to everything else
- ???
- Profit

PR #134, opened 2015-08-28
Cleaned up implementation, added nullability annotations for PFFile.m

nlutsenko:
- Squashed few methods together
- Added nullability annotations for private header
Compliments #133

### DIFF
--- a/Parse/PFFile.m
+++ b/Parse/PFFile.m
@@ -47,7 +47,7 @@ static const unsigned long long PFFileMaxFileSize = 10 * 1024 * 1024; // 10 MB
 
 @implementation PFFile
 
-@synthesize stagedFilePath=_stagedFilePath;
+@synthesize stagedFilePath = _stagedFilePath;
 
 ///--------------------------------------
 #pragma mark - Public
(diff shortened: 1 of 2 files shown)


Parse/PFFile.m
```diff
@@ -315,13 +307,13 @@ - (BFTask *)_uploadFileAsyncWithSessionToken:(NSString *)sessionToken
 #pragma mark Download
 
 - (BFTask *)_getDataAsyncWithProgressBlock:(PFProgressBlock)progressBlock {
-    return [[self _downloadAsyncWithProgressBlock:progressBlock] continueAsyncWithSuccessBlock:^id(BFTask *task) {
+    return [[self _downloadAsyncWithProgressBlock:progressBlock] continueWithSuccessBlock:^id(BFTask *task) {
```
richardjrossiii: Why was this async before? Wasn't it to ensure that all the progress block ticks finished?

nlutsenko: No, not really... It was attached to PFTaskQueue vs PFAsyncTaskQueue, which was safer to use with async APIs.
It's no longer needed. `downloadAsync` already performs on the background thread, so there is basically no need to do this scheduling here.

Issue #110, opened 2015-08-26
Create App Extensions Starter project for iOS/OS X


nlutsenko:
It would be tremendously helpful to have a sample application with at least a single app extension, with all the code inside of it as another starter project.
This would also improve CI testing that our SDKs compile and work for App Extensions.
All of these would be modeled over existing sample apps, and would live just next to them.
- Create iOS ParseAppExtensionsStarterProject in ObjC
- Create iOS ParseAppExtensionsStarterProject in Swift
- Create OS X ParseAppExtensionsStarterProject in ObjC
- Create OS X ParseAppExtensionsStarterProject in Swift
- Add all 4 new starter apps to deployment `rake package:starters`
- Add all 4 new starter apps to Travis-CI via `rake test:starters`

kashif:
@nlutsenko I'll be happy to take this issue if no one else has started it

nlutsenko:
No one did yet, to the best of my knowledge. Feel free to start it!

kashif:
great @nlutsenko will start :dash:

kashif:
@nlutsenko just a couple of quick questions:

* What is the best way to create a new sample application, doing it manually and adding it to the parse workspace? or is there a more automated way?
* which extension type should we go with? I was thinking a Today extension?

nlutsenko:
- You can copy/paste the existing ones and simply rename projects/targets as well as add new targets. Good call on adding this to the workspace.
- Extension types - I think `Today` is a really good example and the most used one. Others, like `Keyboard` or `Share` extension is somewhat hard to showcase, as well as have limitations, e.g. explicitly approved network connections for Keyboard.

Tom-Fox1:
Closing this as no one has reported this issue recently, and we don't have the resources to investigate these historic reports.

Please ask me to reopen if you experience this issue in the latest version of this SDK.

PR #137, opened 2015-08-29
[WIP] Issue #110

kashif:
For issue #110


nlutsenko:
Good start, ping me whenever you are done.
:grinning:

nlutsenko:
@kashif Closing this one out, as it looks it became stale for a long time.
Feel free to reopen or send a new one, since this still would be super useful to have.

kashif:
sure, the problem was that i need to have a developer program in order to turn on the app extension. If i do that then i do not know how it will work when someone else opens the project who does not have a developer program...

nlutsenko:
Do you explicitly need the paid one, or the free will be sufficient? I can test it for you in every case. Let me know how I can help.

PR #139, opened 2015-08-30
Fixed iOS 9 deprecation warnings in PFAlertView, PFPushUtilities.

nlutsenko:
If used with Cocoapods and minimum deployment target is set to iOS 9 - these two places were providing warnings.
- Added compile time check for PFAlertView
- Made PFPushUtilities use PFAlertView

### DIFF
--- a/Parse/Internal/Push/Utilites/PFPushUtilities.m
+++ b/Parse/Internal/Push/Utilites/PFPushUtilities.m
@@ -12,9 +12,9 @@
 #import <dlfcn.h>
 
 #if TARGET_OS_IPHONE
-
 #import <AudioToolbox/AudioToolbox.h>
 
+#import "PFAlertView.h"
 #endif
 
 #import "PFInstallationPrivate.h"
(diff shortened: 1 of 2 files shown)


Parse/Internal/Push/Utilites/PFPushUtilities.m
```diff
@@ -21,6 +21,7 @@
 #import "PFKeychainStore.h"
 #import "PFLogging.h"
 #import "PFMacros.h"
+#import "PFAlertView.h"
```
richardjrossiii: Either wrap this in a `#if TARGET_IOS_IPHONE` check, or change `PFAlertView`'s headers to support OS X. 

I don't have a strong preference to either, though I think that long term making `PFAlertView` cross-platform (with a compile-time warning using NS_AVAILABLE and friends, if you attempt to actually call it on OS X) makes it much simpler to maintain long-term.

code-K:
@nlutsenko with this one i'm getting now an issue... i updated to 1.10.0 through pods...
My target is set to iOS 8

getting now the following..

..../Pods/Parse/Parse/Internal/PFAlertView.h:22:1: 'showAlertWithTitle:message:cancelButtonTitle:otherButtonTitles:completion:' has been explicitly marked unavailable here

..../Pods/Parse/Parse/Internal/Push/Utilites/PFPushUtilities.m:63:18: 'showAlertWithTitle:message:cancelButtonTitle:otherButtonTitles:completion:' is unavailable: not available on iOS (App Extension)


have an extension for today widget.

Thanks much for feedback and best regards,
Christian

nlutsenko:
@code-K, the issue is fixed in #642 and will be included in the new release of the SDK.
You can get it faster - by referncing master of the repo in your Podfile.

code-K:
@nlutsenko Thanks much for the feedback! Works fine with master! :+1:

PR #144, opened 2015-08-31
Improved PFKeyValueCache performance and memory usage

### LINKED ISSUE #141: PFKeyValueCache causing app to be killed because out of memory.
If many cached queries are made, my app spends an incredible amount of time in `-[PFKeyValueCache _compactDiskCache]`. Sometimes between 30s and 1min, and during that time the cpu is running at 100% (out of the 200% available with the 2 cores).

Also memory usage grows really fast during that time (I have yet to determine what is the cause of this). I just ran a test on an iphone 5 with iOS 7, the app quickly reaches about 600Mb and then gets killed by the system (after receiving several memory warnings).

I will look into it more deeply and provide more details and hopefully a way to reproduce. It might take few days since I will be working on an other project this week, but I wanted to report this as soon a possible.

richardjrossiii:
Previously, PFKeyValueCache was re-reading the file attributes from disk for every cached entry, every time we attempted to set a new entry in the cache.

This had serious performance impact for use-cases where repeated caching was used, as copying and recreating the file attributes dictionary caused lots of system calls and wasted memory.

This diff makes PFKeyValueCache much more optimistic, as it does not re-read attributes from disk if the containing folder has not changed. However, this does introduce a potential race-condition that is noted in the code.

Bottom line: This should improve query caching performance for 99% of use-cases.

Fixes #141.

### DIFF
--- a/Parse/Internal/KeyValueCache/PFKeyValueCache.m
+++ b/Parse/Internal/KeyValueCache/PFKeyValueCache.m
@@ -19,6 +19,9 @@
 static const NSUInteger PFKeyValueCacheDefaultDiskCacheSize = 10 << 20;
 static const NSUInteger PFKeyValueCacheDefaultDiskCacheRecords = 1000;
 static const NSUInteger PFKeyValueCacheDefaultMemoryCacheRecordSize = 1 << 20;
+static const NSTimeInterval PFKeyValueCacheDiskCacheTimeResolution = 1; // HFS+ stores only second level accuracy.
+
+static NSString *const PFKeyValueCacheDiskCachePathKey = @"path";
 
 @interface PFKeyValueCacheEntry ()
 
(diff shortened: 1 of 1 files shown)


Parse/Internal/KeyValueCache/PFKeyValueCache.m
```diff
@@ -208,58 +227,109 @@ - (NSString *)_diskCacheEntryForURL:(NSURL *)url {
...
+                       modificationDate:creationDate
+                                   size:bytes.length];
+    } else {
+        [self _invalidateDiskCache];
+    }
 }
 
-- (void)_updateModificationDateAtURL:(NSURL *)url {
-    [self.fileManager setAttributes:@{ NSFileModificationDate: [NSDate date] } ofItemAtPath:url.path error:NULL];
+- (BOOL)_diskCacheDirty {
```
nlutsenko: Nit: `_isDiskCacheDirty`

nlutsenko:
LGTM. Shipit!

PR #145, opened 2015-08-31
Lazy-load PFEventuallyQueue in PFAnalyticsController.

nlutsenko:
- Make `PFAnalyticsController` follow the provider pattern
- Fixed rare main thread stall inside PFAnalyticsController on `trackAppLaunched`
- Fixed rare main thread stall on loading `PFEventuallyQueue`

### DIFF
--- a/Parse/Internal/Analytics/Controller/PFAnalyticsController.m
+++ b/Parse/Internal/Analytics/Controller/PFAnalyticsController.m
@@ -15,6 +15,12 @@
 #import "PFEventuallyQueue.h"
 #import "PFRESTAnalyticsCommand.h"
 
+@interface PFAnalyticsController ()
+
+@property (nonatomic, weak, readonly) PFEventuallyQueue *eventuallyQueue;
+
+@end
+
 @implementation PFAnalyticsController
 
 ///--------------------------------------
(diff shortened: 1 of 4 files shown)


Parse/Internal/Analytics/Controller/PFAnalyticsController.m
```diff
@@ -44,15 +50,19 @@ + (instancetype)controllerWithEventuallyQueue:(PFEventuallyQueue *)eventuallyQue
...
+        // If the Remote Notification payload had a message sent along with it, make
+        // sure to send that along so the server can identify "app opened from push"
+        // instead.
+        id alert = payload[@"aps"][@"alert"];
+        NSString *pushDigest = (alert ? [PFAnalyticsUtilities md5DigestFromPushPayload:alert] : nil);
+
+        PFRESTCommand *command = [PFRESTAnalyticsCommand trackAppOpenedEventCommandWithPushHash:pushDigest
+                                                                                   sessionToken:sessionToken];
+        return [self.eventuallyQueue enqueueCommandInBackground:command];
+    }] continueWithSuccessResult:@YES];
```
nskboy: I wonder why do you need to continue with succes result set to `YES` here?

nlutsenko: Since the result of this operation doesn't matter here, only the error - we are chaining on the task to return `NSNumber numberWithBool:YES` simply to indicate the success of the operation.

`continuWithSuccessResult:` is a simple method wrapper and is identical to
```objc
continueWithSuccessBlock:^id(BFTask *task) {
  return @YES;
}]
```

PR #156, opened 2015-09-01
Fixed rare crash in PFQuery related to weakify/strongify.

richardjrossiii:
In some rare scenarios (e.g. Parse is being torn down), the query will lose all references to it, and the strongify will yield `nil`. When that happens, we cannot read or write to any instance variables in that context.

### DIFF
--- a/Parse/PFQuery.m
+++ b/Parse/PFQuery.m
@@ -842,6 +842,9 @@ static void PFQueryAssertValidOrderingClauseClass(id object) {
                                                                         user:user];
     }] continueWithBlock:^id(BFTask *task) {
         @strongify(self);
+        if (!self) {
+            return task;
+        }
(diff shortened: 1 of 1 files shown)


Parse/PFQuery.m
```diff
@@ -843,7 +843,7 @@ - (BFTask *)_findObjectsAsyncForQueryState:(PFQueryState *)queryState after:(BFT
     }] continueWithBlock:^id(BFTask *task) {
         @strongify(self);
```
nlutsenko: Move the `nil` check here and bail out early, if it's nil, in order not to reference `self` inside `@synchronized()`

nlutsenko:
Single `nit`, otherwise LGTM.
Crazy find.

PR #162, opened 2015-09-03
Make all synchronous APIs have 'throws' in Swift 2.

### LINKED ISSUE #152: Swift v2 Error Handling
Gentlefolk,
 
The Swift v2 try-catch mechanism maps onto Objective-C quite well. Except when there is a shadowing problem.
 
For example:
 
    - (BOOL)pinWithName:(NSString *)name error:(NSError **)error;

 
which automatically maps into Swift v2 in [redacted] v7b6 headers as:
 
    public func pinWithName(name: String, error: ()) throws
 
It should probably map as:

    public func pinWithName(name: String) throws

The Objective-C method Is also commonly paired with this shortened form:
 
    - (BOOL)pinWithName:(NSString *)name;
 
Which maps as: 

    public func pinWithName(name: String) -> Bool

This short method, I suspect, just calls the main one with a NULL for the NSError**. The signature for the short method is remarkably similar to the Swift throwing method. If I leave the short method in my Objective-C header, then errors ensue:

    do { try location.pinWithName(kLocationPin)
        updatePinnedLocations() // Add the locations to the user in the cloud.
    }
    catch let e as NSError { log.error("Local location save error: \(e)") }

Results in the following compile time warning.

    warning: no calls to throwing functions occur within 'try' expression
           do { try location.pinWithName(kLocationPin)
                ^
 
Obviously, the compiler has chosen the wrong signature. In my own code, where I express a similar pattern, I comment out the short method, then the app compiles just fine.

What is the right way to fix the public API? In most things Swift, the answer would be to get rid of the short Objective-C method.
 
Or is there a workaround to the above problem? In my own code, I have tried using the NS_SWIFT_NOTHROWS annotation on the long method signature. It allows me to retain my Swift v1.2 error handling code mostly without change.

What would be the right solution for the Parse API?

Anon,
Andrew

nlutsenko:
- Made all synchronous APIs have `throws` when used from Swift.
- Added backward compatible macro `PF_SWIFT_UNAVAILABLE`, please note that this macro does nothing when compiled with Xcode 6.4, but only works on Xcode 7 (the one that has Swift 2).

After change (Swift 2):
```swift
class PFObject : NSObject {
    public func save() throws
}
```

Before change (Swift 2):
```swift
class PFObject : NSObject {
    public func save() -> Bool
    public func save(error: NSErrorPointer) -> Bool
}
```

Before && After Change (Swift 1.2):
```swift
class PFObject : NSObject {
    public func save() -> Bool
    public func save(error: NSErrorPointer) -> Bool
}
```

closes #152 

cc @richardjrossiii

### DIFF
--- a/Parse/PFCloud.h
+++ b/Parse/PFCloud.h
@@ -28,7 +28,8 @@ PF_ASSUME_NONNULL_BEGIN
 
  @returns The response from the cloud function.
  */
-+ (PF_NULLABLE_S id)callFunction:(NSString *)function withParameters:(PF_NULLABLE NSDictionary *)parameters;
++ (PF_NULLABLE_S id)callFunction:(NSString *)function
+                  withParameters:(PF_NULLABLE NSDictionary *)parameters PF_SWIFT_UNAVAILABLE;
 
 /*!
  @abstract Calls the given cloud function *synchronously* with the parameters provided and
--- a/Parse/PFConfig.h
+++ b/Parse/PFConfig.h
@@ -11,6 +11,7 @@
 
 #import <Bolts/BFTask.h>
 
+#import <Parse/PFConstants.h>
 #import <Parse/PFNullability.h>
 
 PF_ASSUME_NONNULL_BEGIN
@@ -47,7 +48,7 @@ typedef void(^PFConfigResultBlock)(PFConfig *PF_NULLABLE_S config, NSError *PF_N
 
  @returns Instance of `PFConfig` if the operation succeeded, otherwise `nil`.
  */
-+ (PF_NULLABLE PFConfig *)getConfig;
++ (PF_NULLABLE PFConfig *)getConfig PF_SWIFT_UNAVAILABLE;
 
 /*!
  @abstract Gets the `PFConfig` object *synchronously* from the server and sets an error if it occurs.
--- a/Parse/PFConstants.h
+++ b/Parse/PFConstants.h
@@ -420,3 +420,15 @@ typedef void (^PFProgressBlock)(int percentDone);
 #    endif
 #  endif
 #endif
+
+///--------------------------------------
+/// @name Swift Macros
+///--------------------------------------
+
+#ifndef PF_SWIFT_UNAVAILABLE
+#  ifdef NS_SWIFT_UNAVAILABLE
+#    define PF_SWIFT_UNAVAILABLE NS_SWIFT_UNAVAILABLE("")
+#  else
+#    define PF_SWIFT_UNAVAILABLE
+#  endif
+#endif
--- a/Parse/PFFile.h
+++ b/Parse/PFFile.h
@@ -146,7 +146,7 @@ PF_ASSUME_NONNULL_BEGIN
 
  @returns Returns whether the save succeeded.
  */
-- (BOOL)save;
+- (BOOL)save PF_SWIFT_UNAVAILABLE;
 
 /*!
  @abstract Saves the file *synchronously* and sets an error if it occurs.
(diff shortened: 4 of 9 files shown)


grantland:
Looks awesome!

adonoho:
Thank you for doing this. I will import this branch and try it out.

adonoho:
/*!
      @abstract *Synchronously* fetches the PFObject with the current data from the server.
      */
    -- (instancetype)fetch;
    +- (instancetype)fetch PF_SWIFT_UNAVAILABLE;
     /*!
      @abstract *Synchronously* fetches the PFObject with the current data from the server and sets an error if it occurs.
 
      @param error Pointer to an `NSError` that will be set if necessary.
      */
     - (instancetype)fetch:(NSError **)error;
 
     /*!
      @abstract *Synchronously* fetches the `PFObject` data from the server if <isDataAvailable> is `NO`.
      */
    -- (PF_NULLABLE PFObject *)fetchIfNeeded;
    +- (PF_NULLABLE PFObject *)fetchIfNeeded PF_SWIFT_UNAVAILABLE;
 
     /*!
      @abstract *Synchronously* fetches the `PFObject` data from the server if <isDataAvailable> is `NO`.
 
      @param error Pointer to an `NSError` that will be set if necessary.
      */
     - (PF_NULLABLE PFObject *)fetchIfNeeded:(NSError **)error;

I wish to bring your attention the differences between:

     - (instancetype)fetch:(NSError **)error;

and 

     - (PF_NULLABLE PFObject *)fetchIfNeeded:(NSError **)error;

Both are calls that can fail. Forgiving my lack of experience with your framework's patterns,shouldn't they have the same return type that is nullable to ensure that try-catch functions properly? For example:

     - (PF_NULLABLE instancetype)fetch:(NSError **)error;

Should all synchronous methods using try catch return `(PF_NULLABLE instancetype)`?

Anon,
Andrew

nlutsenko:
Thanks for bringing this up! You are absolutely correct, this is our mistake.
All of these methods should have `PF_NULLABLE`, since they will actually return `nil` if there is an error.
Will address in a separate Pull Request.

adonoho:
I suspect you should also consider replacing the `PFObject *` with `instancetype`.

nlutsenko:
:+1:

anthony-lai:
Hi, I had a quick question on how to update my xcode 6.4 code to the xcode 7 version.

Old code:

func a()
    item = b()
    if item == nil {..}
    else {..}

func b() -> PFObject?
    let x = PFObject(className: data)
    let success = x.save()
    if success { return x }
    return nil

now that save doesn't return a boolean, how can I accomplish the same result?

nlutsenko:
@anthony-lai, yes that is correct syntax.
Something like this would fit
```swift
func b() -> PFObject? {
    let x = PFObject(className: "Yolo")
    do {
        try x.save()
        return x
    } catch let error {
        print(error)
        return nil
    }
}
```

You could also make the function itself throw, so you can propagate the error further.
```swift
func b() throws -> PFObject {
    let x = PFObject(className: "Yolo")
    try x.save()
    return x
}
```

richardjrossiii:
@nlutsenko You can also use swift's `try?` syntax, which will discard the error (as parse will already print it for you), and wrap the optional automatically as well. This prevents an additional `do/catch` block :D

anthony-lai:
@nlutsenko @richardjrossiii  Thanks!

nlutsenko:
(－‸ლ) @richardjrossiii you are right... Absolutely forgot about the optional `try?`.
On this note - I still like the `throws` and non-optional functions much more :grin:

PR #166, opened 2015-09-03
Play vibration in PFPush.handlePush() only if default sound is specified.

### LINKED ISSUE #99: [PFPush handlePush:] should not play vibration when not sound is specified.
Currently, sending the userInfo to `[PFPush handlePush:] ` will cause the device to vibrate even if no "sound" parameter is specified in the notification payload, as listed in `PFPush.h #416-422` : 

```
    NSString *soundName = aps[@"sound"];

    if ((id)soundName == [NSNull null] || soundName.length == 0 || [soundName isEqualToString:@"default"]) {
        [[self pushInternalUtilClass] playVibrate];
    } else {
        [[self pushInternalUtilClass] playAudioWithName:soundName];
    }
```

Since is written in the documentation that handlePush: would mimic the behaviour of the notification as if handled by the OS when the app is in background, it would be preferable to not play anything at all if no "sound" parameter is specified.
The drawback would be that it's not possible to play the default notification sound when the parameter is `"sound" : "default"` since there is no way to determine the default system notification sound id (or at least not in a clean-ish way), but at least the developer is not forced to trigger a vibration also for a silent notification with a custom payload.
The code would be something like:

```
   NSString *soundName = aps[@"sound"];     
   if(soundName){
       [[self pushInternalUtilClass] playAudioWithName:soundName];
  }
```

alternatively, default could go to a vibration:

```
    NSString *soundName = aps[@"sound"];
  
       if(soundName){
            if ([soundName isEqualToString:@"default"]) {
                      [[self pushInternalUtilClass] playVibrate];
            } 
           else{
                    [[self pushInternalUtilClass] playAudioWithName:soundName];
           }
  }
```

For further reading:
* http://stackoverflow.com/questions/11776737/how-can-i-access-default-ios-sound-to-set-it-as-notification-sound
* http://stackoverflow.com/questions/9841600/ios-push-notifications-how-to-specify-alert-and-badge-with-no-sound

Hope this has been somehow useful.
Thank you for your great framework :+1:  , best regards

Marco

nlutsenko:
Fixes #99

### DIFF
--- a/Parse/PFPush.m
+++ b/Parse/PFPush.m
@@ -415,12 +415,15 @@ static Class _pushInternalUtilClass = nil;
 
     NSString *soundName = aps[@"sound"];
 
-    if ((id)soundName == [NSNull null] || soundName.length == 0 || [soundName isEqualToString:@"default"]) {
-        [[self pushInternalUtilClass] playVibrate];
-    } else {
-        [[self pushInternalUtilClass] playAudioWithName:soundName];
+    // Vibrate or play sound only if `sound` is specified.
+    if ([soundName isKindOfClass:[NSString class]] && soundName.length != 0) {
+        // Vibrate if the sound is `default`, otherwise - play the sound name.
+        if ([soundName isEqualToString:@"default"]) {
+            [[self pushInternalUtilClass] playVibrate];
+        } else {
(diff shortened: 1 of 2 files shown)


grantland:
LGTM

PR #167, opened 2015-09-03
Cleanup and improve test branch coverage of PFACL.

nlutsenko:
There were few code branches that were not covered - this should be fixed now.
Also - reference `self` or `self class` instead of PFACL directly.


grantland:
LGTM

PR #168, opened 2015-09-03
Added PFFileManager moveItemAsyncAtPath.

nlutsenko:
Implemented an old TODO. We already have `copyItemAsyncAtPath:toPath:` in PFFileManager - so simply add a move method.
Also fixed a bug, where we were returning an NSError from continuation block instead of wrapping with BFTask, so the entire pipe was never failing.


grantland:
LGTM

PR #169, opened 2015-09-03
Replace exceptions in PFUser.signUp with errors.

nlutsenko:
Replace all exceptions with errors, so we don't crash end applications, but rather fail signup requests.
Contributes to #6


grantland:
LGTM, but why does it need to be async?

nlutsenko:
I need a BFTask return type there, so simply named it with `async`, but since I am using default executor inside - it's not really changing threads that much often.

PR #170, opened 2015-09-03
Remove usage of deprecated APIs in PFLocationManager.

nlutsenko:
These APIs are deprecated and no longer needed since we are iOS 7.0+ and OS X 10.9+.


grantland:
LGTM

PR #174, opened 2015-09-04
Fix flaky PurchaseControllerTests.testDownloadAsync test.

nlutsenko:
This test is flaky on Travis because running the command doesn't happen asynchronously, but rather inline (it's mocked to reply with an already completed task) - so sometimes it's completing the overall task faster than it can report progress. Added another expectation for that to make sure that we get a progress callback always.


grantland:
LGTM

PR #181, opened 2015-09-07
Fixed usage of genstrings on El Capitan under Xcode 6.4.

nlutsenko:
For whatever reason `genstrings` is not available in Xcode 6.4 on El Capitan.
Update to use `extractLocStrings` which looks like the same thing.
Tested locally by removing the generated localization.


richardjrossiii:
Genstrings is still there, it just needs to be through XCRun. I don't have a huge preference between `xcrun extractLocStrings ` and `xcrun genstrings`, so this is fine for now. Shipit

PR #196, opened 2015-09-08
Enable address & runtime sanitizers for debug & test builds.

richardjrossiii:
These sanitizers (documented here: http://clang.llvm.org/docs/UsersManual.html#controlling-code-generation) will help us catch undefined behavior while unit testing, and not at run-time, improving the overall stability of the SDKs.


nlutsenko:
Looks great. Please move these into `Test.xcconfig`, since you want the sanitizer to be used when testing, right?

PR #197, opened 2015-09-09
Add https support for PFFiles.

### LINKED ISSUE #12: PFFile TLS Download Support
iOS 9 and OS X 10.11 has a requirement that any network connection, unless added to exceptions is done via TLS. [Apple Tech Note](https://developer.apple.com/library/prerelease/ios/technotes/App-Transport-Security-Technote/).
We have full support right now for API server, as announced in the [blog post](http://blog.parse.com/announcements/moving-parse-apps-to-the-more-secure-sha-2-industry-standard/).

The only piece left is file downloads, which are still being served from S3 via `http` and we are working on getting support for `TLS` for them shortly.

The current workaround for this issue is to add the following section to `Info.plist` of your app:
```
<key>NSExceptionDomains</key>
<dict>
    <key>files.parsetfss.com</key>
    <dict>
        <key>NSTemporaryExceptionAllowsInsecureHTTPLoads</key>
        <true/>
    </dict>
</dict>
```

Please comment on this issue if there is anything we can help you with regarding PFFile TLS support.

nlutsenko:
This is still pending backend changes to actually enable https, but just throwing this in, since this is the only client-side change required.
Fixes #12

### DIFF
--- a/Parse/Internal/File/State/PFFileState.m
+++ b/Parse/Internal/File/State/PFFileState.m
@@ -13,6 +13,14 @@
 #import "PFMutableFileState.h"
 #import "PFPropertyInfo.h"
 
+static NSString *const _PFFileStateSecureDomain = @"files.parsetfss.com";
+
+@interface PFFileState ()
+
+@property (nonatomic, copy, readwrite) NSString *secureURLString;
+
+@end
+
 @implementation PFFileState
 
 ///--------------------------------------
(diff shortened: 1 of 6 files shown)


Parse/Internal/File/State/PFFileState.m
```diff
@@ -47,6 +55,43 @@ - (instancetype)initWithName:(NSString *)name urlString:(NSString *)urlString mi
...
+    }
+}
+
+- (NSString *)secureURLString {
+    if (_secureURLString) {
+        return _secureURLString;
+    }
+
+    if (!self.urlString) {
+        return self.urlString;
```
richardjrossiii: Change this to `return nil` for readability?

nlutsenko: :+1:

PR #198, opened 2015-09-09
Added response body string to network notifications, don't supply any user info if running outside of debug mode.

nlutsenko:
- Added another property on every PFURLSessionDataTaskDelegate - `responseString`, which is set only if we have a string to set there
- Added new user info key in PFCommandRunningConstants
- Use `PFLogger.sharedLogger()` to get the log level
- Add user info only if logLevel is set to `PFLogLevelDebug`

cc @wangmengyan95 @grantland

### DIFF
--- a/Parse/Internal/Commands/CommandRunner/PFCommandRunningConstants.h
+++ b/Parse/Internal/Commands/CommandRunner/PFCommandRunningConstants.h
@@ -50,10 +50,18 @@ extern NSString *const PFCommandRunnerDidReceiveURLResponseNotification;
 
 /*!
  @abstract The key of request(NSURLRequest) in the userInfo dictionary of a notification.
+ @note This key is populated in userInfo, only of `PFLogLevel` in `PFLogger` is set to `PFLogLevelDebug`.
  */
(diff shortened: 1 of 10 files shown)


Parse/Internal/Commands/CommandRunner/PFCommandRunningConstants.h
```diff
@@ -57,3 +57,8 @@ extern NSString *const PFCommandRunnerNotificationURLRequestUserInfoKey;
  @abstract The key of response(NSHTTPURLResponse) in the userInfo dictionary of a notification.
  */
 extern NSString *const PFCommandRunnerNotificationURLResponseUserInfoKey;
+
+/*!
+ @abstract The key of repsonse body (usually `NSString` with JSON) in the userInfo dictionary of a notification.
```
richardjrossiii: Document that it's only available when the logLevel is debug?

nlutsenko: Yup, sounds like a good idea...

nlutsenko: Done, for all `userInfo` key constants.

PR #201, opened 2015-09-09
Support Bolts generics in internal headers, part 2.

richardjrossiii:
Part 2/5.

Adds definitions to the following internal sections:
 - ACL
 - Analytics
 - CloudCode
 - Commands

Each should be able to function completely independent of the others.

### DIFF
--- a/Parse/Internal/CloudCode/PFCloudCodeController.h
+++ b/Parse/Internal/CloudCode/PFCloudCodeController.h
@@ -40,8 +40,8 @@
 
  @returns `BFTask` with a result set to a result of Cloud Function.
  */
-- (BFTask *)callCloudCodeFunctionAsync:(NSString *)functionName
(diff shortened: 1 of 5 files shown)


Parse/Internal/CloudCode/PFCloudCodeController.h
```diff
@@ -40,8 +40,8 @@
 
  @returns `BFTask` with a result set to a result of Cloud Function.
  */
-- (BFTask *)callCloudCodeFunctionAsync:(NSString *)functionName
-                        withParameters:(NSDictionary *)parameters
-                          sessionToken:(NSString *)sessionToken;
+- (BFTask PF_GENERIC(id) *)callCloudCodeFunctionAsync:(NSString *)functionName
```
nlutsenko: Nit: Not sure if you need it - since `id` is inferred by default.

PR #207, opened 2015-09-09
Update bitcode support for Xcode 7 GM.

nlutsenko:
User other c flags, make sure that everything works great on both Xcode 6.4 and Xcode 7.0.


the-scott-davis:
@nlutsenko this is still an issue with Xcode 7 official release running against iOS 9 base SDK.... still throwing the enable bitcode message

nlutsenko:
Discussing on #254

PR #209, opened 2015-09-09
Fix a few minor undefined behavior issues.

richardjrossiii:
One could have resulted in a stack smash, and the other would cause issues if we ever linked with C++ anywhere.


nlutsenko:
LGTM

PR #212, opened 2015-09-10
Added private method for checking if user is linked with a given auth type.

nlutsenko:
- Add new private method
- Update `PFAnonymousUtils` to use it

cc @grantland


grantland:
LGTM

PR #218, opened 2015-09-10
Base work for public Bolts task generics.

richardjrossiii:
We need to be able to use PF_GENERIC with multiple types (e.g. `NSDictionary`). Make it varargs.

### DIFF
--- a/Parse/PFConstants.h
+++ b/Parse/PFConstants.h
@@ -438,8 +438,8 @@ typedef void (^PFProgressBlock)(int percentDone);
 ///--------------------------------------
 
 #if __has_feature(objc_generics) || __has_extension(objc_generics)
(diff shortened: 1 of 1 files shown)


Parse/PFConstants.h
```diff
@@ -438,8 +438,10 @@ typedef void (^PFProgressBlock)(int percentDone);
...
 #if __has_feature(objc_generics) || __has_extension(objc_generics)
-#  define PF_GENERIC(type) <type>
+#  define PF_GENERIC(...) <__VA_ARGS__>
 #else
-#  define PF_GENERIC(type)
+#  define PF_GENERIC(...)
 #  define PFGenericObject PFObject *
 #endif
+
+
```
nlutsenko: Kill whitespace.

PR #223, opened 2015-09-10
Added BFTask generics to header of PFCloud

richardjrossiii:
Bolting down clouds. Is that illegal?


nlutsenko:
You don't need this, it's defaulted to `id`.

PR #225, opened 2015-09-10
Added BFTask generics to header of PFConfig

richardjrossiii:
Configure bolts.


nlutsenko:
I can see a pattern here...

PR #227, opened 2015-09-10
Added BFTask generics to header of PFObject

richardjrossiii:
This one actually has content!

### DIFF
--- a/Parse/PFObject.h
+++ b/Parse/PFObject.h
@@ -210,7 +210,7 @@ NS_REQUIRES_PROPERTY_DEFINITIONS
 /*!
  @abstract Clears any changes to this object's key that were done after last successful save and sets it back to the
  server state.
- 
+
  @param key The key to revert changes for.
  */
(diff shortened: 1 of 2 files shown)


Parse/PFObject.h
```diff
@@ -569,7 +569,7 @@ NS_REQUIRES_PROPERTY_DEFINITIONS
 
  @returns The task that encapsulates the work being done.
  */
-- (BFTask *)fetchInBackground;
+- (BFTask PF_GENERIC(__kindof PFObject *)*)fetchInBackground;
```
nlutsenko: Can we use `instancetype`? It's ok if not - I am just wondering if it will let us do that, since this is purely `BFTask` incapsulating `self`.

nlutsenko:
Single question, but LGTM.

Parse/PFObject.h
richardjrossiii: We cannot. Instancetype can only be used as the **complete** return type of a method (e.g. `const instancetype` is also invalid).

PR #236, opened 2015-09-11
Make PFUser.unlinkWithAuthTypeInBackground() fully async.

nlutsenko:
Implementing a TODO.

### DIFF
--- a/Parse/PFUser.m
+++ b/Parse/PFUser.m
@@ -877,18 +877,16 @@ static BOOL revocableSessionEnabled_;
 #pragma mark Unlink
 
 - (BFTask *)unlinkWithAuthTypeInBackground:(NSString *)authType {
-    // TODO: (nlutsenko) Make it fully async.
-    BFTask *save = nil;
-    @synchronized ([self lock]) {
-        if (!self.authData[authType]) {
-            save = [BFTask taskWithResult:@YES];
-        } else {
-            self.authData[authType] = [NSNull null];
-            dirty = YES;
-            save = [self saveInBackground];
+    return [BFTask taskFromExecutor:[BFExecutor defaultPriorityBackgroundExecutor] withBlock:^id{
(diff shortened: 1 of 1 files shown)


Parse/PFUser.m
```diff
@@ -877,18 +877,19 @@ - (BFTask *)linkWithAuthTypeInBackground:(NSString *)authType authData:(NSDictio
...
-            save = [BFTask taskWithResult:@YES];
-        } else {
-            self.authData[authType] = [NSNull null];
-            dirty = YES;
-            save = [self saveInBackground];
+    return [BFTask taskFromExecutor:[BFExecutor defaultPriorityBackgroundExecutor] withBlock:^id{
+        BFTask *task = nil;
+        @synchronized ([self lock]) {
+            if (!self.authData[authType]) {
+                task = [BFTask taskWithResult:@YES];
```
richardjrossiii: We can just return from here, why are we putting it into a task variable?

nlutsenko: Good point. Will change.

nlutsenko:
Updated.

Parse/PFUser.m
```diff
@@ -877,18 +877,16 @@ - (BFTask *)linkWithAuthTypeInBackground:(NSString *)authType authData:(NSDictio
...
+        @synchronized (self.lock) {
+            if (self.authData[authType]) {
+                self.authData[authType] = [NSNull null];
+                dirty = YES;
+            }
+            return [self saveInBackground];
         }
-    }
-    return save;
+        return @YES;
```
richardjrossiii: I think this logic is wrong. this is unreachable code, and we should only save if we changed the auth data.

nlutsenko:
Right... Good catch. Too much logic for Friday...

nlutsenko:
Updated and ready.

PR #239, opened 2015-09-11
Fixed an issue where LDS wouldn't search the whole subtree for objects not yet fetched.

### LINKED ISSUE #131: Local Datastore losing information for PFObjects when app is killed
In an app I'm working on, I have a class named Group.  It subclasses PFObject, and has two variables: owner (which is a reference to the current user) and name, which is a String. I have Two scenarios, same code base - 

Scenario 1:
1. Log into app
2. Put phone in airplane mode
3. Create 4 groups with the following names: a, aa, aaa, aaaa
4. Kill the app
5. Open the app
6. Go to groups page - only one group is listed, with name aaaa
7. Take phone out of airplane mode
8. Wait 10-15 seconds (long enough for phone to connect and have enough time to save to the server)
9. Refresh group list, still only see group aaaa

Scenario 2:
1. Log into app
2. Put phone in airplane mode
3. Create 4 groups with the following names: b, bb, bbb, bbbb
4. Take phone out of airplane mode (we have not killed the app at this point, but it’s important to note that backgrounding/foregrounding the app has no effect or issue)
5. Wait 10-15 seconds (long enough for phone to connect and have enough time to save to the server
6. Refresh group list, see all four groups still exist

When I look online at our Group table, I see eight rows.  Five of these rows have all the correct info - the rows for aaaa, b, bb, bbb, bbbb.  The other 3 rows (assumedly for the groups I named a, aa, and aaa) have info for objectId, createdAt, updatedAt, and ACL - but name and owner are both listed as undefined.  This leads me to believe that the local datastore is losing that info when the app is killed.  Here is my code that is giving the issue:

Group class:
```
class Group: PFObject, PFSubclassing {
    
    static func parseClassName() -> String {
        return "Group"
    }
    
    @NSManaged var owner: TrustTreeUser // This class subclasses PFUser
    @NSManaged var name: String
    
    func saveGroup() {
        if IJReachability.isConnectedToNetwork() {
            self.saveInBackground()
        } else {
            self.saveEventually()
        }
    }
}
```

Creating and saving the Group instance:
```
var newGroup = Group()
newGroup.name = groupName // this is a string passed in from a UIAlertController
newGroup.owner = TrustTreeUser.currentUser()!
newGroup.saveGroup()
```

This is is happening running iOS 8.4.1, Xcode 6.4, and has been experienced in Parse versions 1.8.0-1.8.2

richardjrossiii:
This caused issues when using the eventually queue, as the eventually pin objects were getting discarded immediately after creation, and never re-fetched from the database afterward.

Fixes #131.

cc @grantland Does this same issue exist on android?

### DIFF
--- a/Parse/Internal/PFEventuallyPin.m
+++ b/Parse/Internal/PFEventuallyPin.m
@@ -158,7 +158,14 @@ NSString *const PFEventuallyPinKeyCommand = @"command";
         pin[PFEventuallyPinKeyCommand] = commandDictionary;
     }
 
-    return [[pin pinInBackgroundWithName:PFEventuallyPinPinName] continueWithBlock:^id(BFTask *task) {
+    // NOTE: This is needed otherwise ARC releases the pins before we have a chance to persist the new ones to disk,
+    // Which means we'd lose any columns on objects in eventually pins not currently in memory.
+    __block NSArray *existingPins = nil;
+    return [[[self findAllEventuallyPin] continueWithSuccessBlock:^id(BFTask *task) {
+        existingPins = task.result;
+        return [pin pinInBackgroundWithName:PFEventuallyPinPinName];
+    }] continueWithSuccessBlock:^id(BFTask *task) {
+        existingPins = nil;
         return pin;
     }];
 }


nlutsenko:
Something around `includeKey` for queries for Eventually Pins instead of this change to save memory.

richardjrossiii:
Changed to be fixed in eventually pin, rather than offlinestore.

nlutsenko:
Fabulous!

PR #241, opened 2015-09-12
Added PFFileDataStream for streaming file Downloads.

richardjrossiii:
This is an NSInputStream proxy that is used for file downloads while concurrently reading and writing to a file. This allows us to truly stream from parse using `-[PFFile getDataDownloadStreamInBackground]`, whereas before it had the potential to break if you consumed it too fast.

### DIFF
--- a/Parse.xcodeproj/project.pbxproj
+++ b/Parse.xcodeproj/project.pbxproj
@@ -1138,6 +1138,8 @@
 		F55C740D1B631557000EDAFA /* PFURLSessionCommandRunner_Private.h in Headers */ = {isa = PBXBuildFile; fileRef = F55C740B1B631557000EDAFA /* PFURLSessionCommandRunner_Private.h */; };
 		F5732DE11B6712140066DCD5 /* URLSessionDataTaskDelegateTests.m in Sources */ = {isa = PBXBuildFile; fileRef = F5732DE01B6712140066DCD5 /* URLSessionDataTaskDelegateTests.m */; };
 		F5732DE21B6712140066DCD5 /* URLSessionDataTaskDelegateTests.m in Sources */ = {isa = PBXBuildFile; fileRef = F5732DE01B6712140066DCD5 /* URLSessionDataTaskDelegateTests.m */; };
+		F57E29B21BA388DD00A2C59D /* FileDataStreamTests.m in Sources */ = {isa = PBXBuildFile; fileRef = F57E29B11BA388DD00A2C59D /* FileDataStreamTests.m */; };
+		F57E29B31BA388DD00A2C59D /* FileDataStreamTests.m in Sources */ = {isa = PBXBuildFile; fileRef = F57E29B11BA388DD00A2C59D /* FileDataStreamTests.m */; };
 		F586B3511B1E3BD70082E3BD /* PFBaseState.m in Sources */ = {isa = PBXBuildFile; fileRef = F586B34F1B1E3BD70082E3BD /* PFBaseState.m */; };
 		F586B3521B1E3BE90082E3BD /* PFBaseState.m in Sources */ = {isa = PBXBuildFile; fileRef = F586B34F1B1E3BD70082E3BD /* PFBaseState.m */; };
 		F589894B1B7427FF008A566B /* AlertViewTests.m in Sources */ = {isa = PBXBuildFile; fileRef = 814915CE1B66D44500EFD14F /* AlertViewTests.m */; };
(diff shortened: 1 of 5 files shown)


Parse/Internal/File/NikitaInputStream/NikitaInputStream.h
```diff
@@ -0,0 +1,31 @@
...
+ * of patent rights can be found in the PATENTS file in the same directory.
+ */
+
+#import <Foundation/Foundation.h>
+
+NS_ASSUME_NONNULL_BEGIN
+
+/*!
+ @abstract Why? Because Nikita.
+ @discussion Actually, this does have some practical use. When downloading a file stream via
```
nlutsenko: I don't remember precisely, but I think these annotations are simply ignored for class documentation.

Parse/Internal/File/NikitaInputStream/NikitaInputStream.m
```diff
@@ -0,0 +1,105 @@
...
+#import "NikitaInputStream.h"
+
+#import <objc/runtime.h>
+#import <objc/message.h>
+
+@interface NikitaInputStream() <NSStreamDelegate>
+@end
+
+@implementation NikitaInputStream {
+    NSString *_path;
```
nlutsenko: Nit: Move all instance variables into class extension above to be consistent in code style.

Parse/Internal/File/NikitaInputStream/NikitaInputStream.m
```diff
@@ -0,0 +1,105 @@
...
+    return self;
+}
+
+- (void)stopBlocking {
+    _finished = YES;
+
+    [self stream:_inputStream handleEvent:NSStreamEventHasBytesAvailable];
+}
+
+#pragma mark - NSProxy methods
```
nlutsenko: Nit: Use common pragma mark, aka
```objc
#pragma mark - NSProxy Methods
```

Parse/Internal/File/NikitaInputStream/NikitaInputStream.m
```diff
@@ -0,0 +1,105 @@
...
+- (void)forwardInvocation:(NSInvocation *)invocation {
+    [invocation invokeWithTarget:_inputStream];
+}
+
+- (BOOL)respondsToSelector:(SEL)aSelector {
+    Method implementation = class_getInstanceMethod([self class], aSelector);
+    return implementation ? YES : [_inputStream respondsToSelector:aSelector];
+}
+
+#pragma mark - NSInputStream methods
```
nlutsenko: Nit: Same as line 44.

Tests/Unit/NikitaInputStreamTests.m
```diff
@@ -0,0 +1,106 @@
+//
+//  NikitaInputStreamTests.m
```
nlutsenko: Nit: Add proper common copyright.

Parse/Internal/File/NikitaInputStream/NikitaInputStream.m
```diff
@@ -0,0 +1,105 @@
...
+    // Reading 0 bytes from an NSInputStream causes this strange undocumented behavior: it marks the stream as 'at end',
+    // regardless of whether more bytes are available or not. lolwut?
+    if (len == 0) {
+        return 0;
+    }
+
+    return [_inputStream read:buffer maxLength:len];
+}
+
+#pragma mark - NSStreamDelegate
```
nlutsenko: Nit: Same as line 44.

nlutsenko:
Looks great, minus nits.
Don't really have a good name for this class, sorry... You can name it anything, but please prefix with `PF`.

The only question about this that I have - what is the expected behavior for using this stream with `getBuffer:length:`? It will simply call through into the actual stream, just wondering what are the things people can run into.

richardjrossiii:
`getBuffer:length:` is not implemented (and probably will never be) for the built-in `NSInputStream` classes, and the docs aren't clear how it should be used in any case (e.g. whether this buffer already has contents, or whether you should call `-read:maxLength:` after getting the buffer).

According to [this](http://www.cocoabuilder.com/archive/cocoa/100909-nsinputstream-getbuffer-length.html) mailing list it would appear that you should still call `-read:maxLength:`, but according to [this](http://lists.apple.com/archives/cocoa-dev/2012/Jul/msg00826.html), it would appear you should not.

Seeing as apple has never used this themselves and probably never will due to the API confusion surrounding it, I think it's safe enough to ignore this for now.

nlutsenko:
Sounds great! Fix nits and let's ship!

nlutsenko:
Please rename before shipping. As much as I love the name (it has my name in it :D), not sure if other people will understand this.

PR #252, opened 2015-09-14
Cleanup third party authentication methods in PFUser.

nlutsenko:
- Renamed authentication provider to `PFUserAuthenticationDelegate`
- Cleaned up that protocol
- Updated documentation and methods in PFUser
- Updated tests and implementation

cc @grantland


Parse/PFUserAuthenticationDelegate.h
```diff
@@ -0,0 +1,38 @@
...
+/*!
+ Provides a general interface for delegation of third party authentication with <PFUser>s.
+ */
+@protocol PFUserAuthenticationDelegate <NSObject>
+
+/*!
+ @abstract Called when restoring third party authentication credentials that have been serialized, 
+ such as session keys, user id, etc.
+ 
+ @note This method will be execute on a background thread.
```
grantland: s/execute/executed/g

grantland:
I don't think we need `PFAnonymousAuthenticationProvider` anymore, since it doesn't do anything and we already return `true` if the delegate doesn't exist.

Other than that and a small nit, LGTM!

nlutsenko:
Discussed offline, but putting here for reference: We still need it, since we are checking for it when logging in / linking, as well as we don't really want to allow overriding of anonymous authentication source.

PR #255, opened 2015-09-14
Fixed crash when using PFObject.deleteAll() with empty array.

### LINKED ISSUE #253: Calling deleteAll:error: on an empty array causes a crash
This will cause a crash :
```
dispatch_async(dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_BACKGROUND, 0), ^{
	[PFObject deleteAll:@[] error:nil];
});
```
This is because `deleteAllInBackground:` returns a task whose result is not a boolean wrapped in an `NSNumber` when the array is empty.

nlutsenko:
- Fixed wrong return type for `deleteInBackground:` if objects array is empty.
- Added tests for all of this.
Fixes #253

### DIFF
--- a/Tests/Unit/ObjectUnitTests.m
+++ b/Tests/Unit/ObjectUnitTests.m
@@ -11,6 +11,7 @@
 #import "PFUnitTestCase.h"
 #import "Parse_Private.h"
 #import "PFObjectPrivate.h"
+#import "BFTask+Private.h"
 
 @interface ObjectUnitTests : PFUnitTestCase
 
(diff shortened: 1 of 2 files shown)


Tests/Unit/ObjectUnitTests.m
```diff
@@ -199,6 +200,30 @@ - (void)testFetchObjectWithoutObjectIdError {
...
+    XCTAssertTrue([PFObject deleteAll:nil]);
+    XCTAssertTrue([PFObject deleteAll:@[]]);
+
+    NSError *error = nil;
+    XCTAssertTrue([PFObject deleteAll:nil error:&error]);
+    XCTAssertNil(error);
+    XCTAssertTrue([PFObject deleteAll:@[] error:&error]);
+    XCTAssertNil(error);
+
+    XCTAssertTrue([[PFObject deleteAllInBackground:nil] waitForResult:nil]);
```
richardjrossiii: Don't use `XCTAssertTrue()` for object types.

nlutsenko: (－‸ლ)

PR #267, opened 2015-09-16
Fixed PFInstallation not updating automatic fields when using PFObject.saveAll().

### LINKED ISSUE #261: PFInstallation specific properties not set when calling [PFObject saveAllInBackground block]
My app sometimes saves an array of objects to Parse using [PFObject saveAllInBackground:objectsToSave block:^(BOOL succeeded, NSError *error)

This has been working fine, but since maybe a week or two some of the PFInstallation specific properties are not set. appIdentifier, appName, appVersion, badge, parseVersion, timeZone, localeIdentifier are all 'undefined'. All other properties are saved as expected. 

I solved this by calling [[PFInstallation currentDevice] saveInBackgroundWithBlock:^(BOOL succeeded, NSError *error) and now it works.

But this seems to be a bug. I tried rolling back to older versions of Parse iOS SDK but no change. So, possibly this is an issue server-side.

Thanks! Keep up the great work you are doing over at Parse :-)

nlutsenko:
Not coverable with unit tests just yet...

Fixes #261

### DIFF
--- a/Parse/Internal/Object/PFObjectPrivate.h
+++ b/Parse/Internal/Object/PFObjectPrivate.h
@@ -38,6 +38,8 @@
 
 @protocol PFObjectPrivateSubclass <NSObject>
 
+@required
+
 ///--------------------------------------
 /// @name State
 ///--------------------------------------
@@ -57,6 +59,18 @@
  @returns Task that encapsulates the validtion.
  */
 - (BFTask *)_validateSaveEventuallyAsync;
+
+@optional
+
+///--------------------------------------
+/// @name Before Save
+///--------------------------------------
+
+/*!
+ Called before an object is going to be saved. Called in a context of object lock.
+ Subclasses can override this method to do any custom updates before an object gets saved.
+ */
+- (void)_objectWillSave;
 
 @end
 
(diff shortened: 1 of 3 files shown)


Parse/Internal/Object/PFObjectPrivate.h
```diff
@@ -58,6 +60,18 @@
...
 - (BFTask *)_validateSaveEventuallyAsync;
 
+@optional
+
+///--------------------------------------
+/// @name Before Save
+///--------------------------------------
+
+/*!
+ Called before an object is going to be safe. Called in a context of object lock.
```
richardjrossiii: s/safe/saved/g.

Parse/PFInstallation.m
```diff
@@ -257,7 +242,7 @@ - (BOOL)needsDefaultACL {
 #pragma mark - Automatic Info
 ///--------------------------------------
 
-- (void)_updateAutomaticInfo {
+- (void)_objectWillSave {
     if ([self _isCurrentInstallation]) {
         @synchronized(self.lock) {
```
richardjrossiii: We already have the lock here, no need for it.

PR #272, opened 2015-09-16
Fix crash related to nil completion handler.

### LINKED ISSUE #269: Parse SDK crasheson 'ok'-click on versions higher than 1.8.2
When using Parse SDK 1.8.3 or higher the app crashes when pressing "ok" on the alertView.

The method showAlertWithTitle:message:etc. crashes because of a nil value as the 'completion' argument.

richardjrossiii:
Super simple bug, super simple fix.

Fixes #269.


nlutsenko:
:facepalm:

PR #281, opened 2015-09-17
Enable support for bitcode and ios9

dgrekov:
Converted the build to a dynamic framework, enabled bitcode, unfortunately, could not figure out how to make Bolts build automatically.


nlutsenko:
Amazing! This looks like a great start!
One of the things that is very important for us is that we want to make sure that iOS 7.0 still is supported os version and as a result - Static Framework on iOS probably still need to be there.

dgrekov:
This references issue #8 by the way.

Agree with the challenge of supporting ios7, however it seems that static frameworks will not allow the app to compile with bitcode enabled, will explore further and will update this PR as I can identify more details.

dgrekov:
I will also look to revise the PR based on the above comments, thank you for your feedback.

nlutsenko:
Actually, we have bitcode support using static frameworks for SDK versions starting from 1.8.3.

PR #289, opened 2015-09-18
Use NSURLComponents to construct URLs instead of CFURL.

### LINKED ISSUE #284: Parse 1.8.5 throws deprecation warning on iOS 9
This warning is being thrown on iOS 9, XCode 7 and Swift 2.

'CFURLCreateStringByAddingPercentEscapes' is deprecated: first deprecated in iOS 9.0 - Use [NSString stringByAddingPercentEncodingWithAllowedCharacters:] instead, which always uses the recommended UTF-8 encoding, and which encodes for a specific URL component or subcomponent (since each URL component or subcomponent has different rules for what characters are valid).

nlutsenko:
Since we are not targeting iOS <7.0 and OS X <10.9 - use `NSURLComponents` instead of our own implementation of that.
Fixes #284


richardjrossiii:
burninated.

PR #293, opened 2015-09-18
Added an ability to download a file and get a file path to PFFiles.

nlutsenko:
- New public API
- Updates tests
- A little bit of cleanup

cc @grantland @wangmengyan95


nlutsenko:
Updated naming, added few notes in documentation.

PR #294, opened 2015-09-19
Update all NSNull generic types to use NSNumber instead.

nlutsenko:
- `deleteEventually` return NSNumber*
- `enableRevocableSessions` and `logOutInBackground` now don't specify a return type since they are not useful at all (the contract there is to yield `nil`)

cc @grantland


nlutsenko:
Woohoo!

Issue #299, opened 2015-09-20
Deadlock on background thread running query


yoavramov:
Hi,

I faced a deadlock on 5 background threads which got stocked on @synchronized(lock) when running parse queries. I didn't run at the same moment all the 5 queries, but since the user hasn't received a response, the user left the page and return it (which cause the query to run once again).

This is not the first time I faced a deadlock in Parse SDK.
I also find out that calling to PFUser.currentUser() can cause a deadlock on main thread because BFTask use semaphore with "wait for ever"
One more deadlock I faced is which accessing a Parse object from the main thread to read one of its properties while a background thread is updating the object from a query result

The two last deadlocks I've mentioned are locking the main thread, therefore the app is freeze and after a short time the OS terminating it.
The first deadlock is only on background thread, so the app isn't freeze, but the user won't receive any data from server.

nlutsenko:
Hey @yoavramov, thanks for the report!
Any chance you can post the stack trace for this issue?
It's really hard to debug these as the flow you are describing works great, and I suspect it's closely related to your schema/flow of execution.

yoavramov:
<Thread 1: MainThread>(
2   MyApp                           	 0x00b4bc05 -[PFOfflineStore getOrCreateObjectWithoutDataWithClassName:objectId:] + 316,
3   MyApp                           	 0x00b1d429 +[PFObject objectWithoutDataWithClassName:objectId:] + 208,
4   MyApp                           	 0x00b1d735 +[PFObject objectWithoutDataWithObjectId:] + 216,
5   MyApp                           	 0x0032c868 PFUser.withoutDataWithObjectId + 220,
6   MyApp                           	 0x0031cd70 <private> + 124,
7   MyApp                           	 0x0064728c <private> + 1656,
8   MyApp                           	 0x0064819c <private> + 168,
9   MyApp                           	 0x005a5280 <private> + 408,
10  MyApp                           	 0x005a2800 <private> + 340,
11  MyApp                           	 0x005a2eac <private> + 56,
12  UIKit                               0x2e56855d <redacted> + 600,
13  UIKit                               0x2e5682cd <redacted> + 24,
14  UIKit                               0x2e6f1619 <redacted> + 628,
15  UIKit                               0x2e61242b <redacted> + 422,
16  UIKit                               0x2e612235 <redacted> + 44,
17  UIKit                               0x2e6121c9 <redacted> + 184,
[...]

Issue #61, opened 2015-08-19
Accessing currentUser on multiple threads causing deadlock


eeallen1:
When I'm launching my app, there are several places where I'm using PFUser.CurrentUser - mostly for queries, but also a couple of writes. It looks like the deadlock occurs when one thread is setting an object on currentUser while another is using it in a query, but I'm not entirely sure. 

Here is a stack trace:
![screen shot 2015-08-19 at 4 13 08 pm](https://cloud.githubusercontent.com/assets/6464213/9369175/1ea8d512-468d-11e5-858f-128f903a50a1.png)

I know that there have been similar issues opened and resolved for other releases, and it may be that 1.8 has nothing to do with it, but it only started happening in the last couple of days (after I updated). 

Some more details: 

- Deadlock doesn't occur every launch, and it appears to happen much more frequently on older devices (iphone 4s) than the newer models. It almost never occurs on the simulator. 

- I've tried commenting out dozens of lines of code that the stack trace points to, but it always end up getting hung somewhere else. I haven't procedurally uncommented and tested every call to currentUser because so much code depends on it, but that'll be my next step for troubleshooting. 

**Edit**
Is it possible that the deadlock is occurring during Facebook authentication while I'm trying to query/modify currentUser elsewhere? Take a look at this: 

![screen shot 2015-08-19 at 4 30 21 pm](https://cloud.githubusercontent.com/assets/6464213/9369573/c34da53c-468f-11e5-8828-f19180319965.png)

nlutsenko:
Hey @eeallen1, thanks for reporting this.
I've tried to reproduce the problem, but it doesn't repro for me, I am going to try investigating it further and there is a single piece that I think the bug must be hiding behind.
I'll update this issue once I have more information or attach a PR to this.

eeallen1:
@nlutsenko If there's anything specific I can look for that would help you, let me know. I tried creating an "empty" version of my app with the same number of Parse requests, but I can't seem to recreate the error when I take out the processor-intensive parts of the app. 

I don't know much about how multithreading works, but I'm guessing this issue only arises when the processor can't switch quickly enough between threads. I could be totally off base, but the deadlock does occur significantly less frequently on the simulator and the newest iPhone than on the iPhone 4s and 5.

kross50:
[...]

PR #306, opened 2015-09-23
    Fixed deadlock in -collectDirtyChildren:

richardjrossiii:
By no longer holding a lock while recursing, we allow other threads who may be waiting on the current object to progress before iterating to children of the object, solving some of the deadlocks that we've seen in issues #11, #61, and #299.

This does not necessarily fix the above issues, as we probably still have similar deadlocks elsewhere in the codebase.

This is based on the mutable containers branch until the time comes to land this, as we want some of our users to be able to test this, and mutable containers also contains deadlocks.

cc @grantland

### DIFF
--- a/Parse/PFObject.m
+++ b/Parse/PFObject.m
@@ -263,6 +263,7 @@ static BOOL PFObjectValueIsKindOfMutableContainerClass(id object) {
 
     } else if ([node isKindOfClass:[PFObject class]]) {
         PFObject *object = (PFObject *)node;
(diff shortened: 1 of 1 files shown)


nlutsenko:
Rebate on master?

richardjrossiii:
Get your rebates folks! Parse, now 50% off! :grin: :dollar: 

Like I said in the description. Keeping this one on mutable containers so we can have some of our users more easily test it.

nlutsenko:
LGTM
Merge after rebasing.

PR #308, opened 2015-09-23
Fix copy/paste error in retrieving userController.

swehrli:
I run into an issue where then `becomeInBackground` function returned neither a user object nor an error after a login/logout cycle. I tracked the error down to the `PFUserController` returning `nil` for the `userController` because of what seems to be a copy/paste error when checking for an existing controller.


richardjrossiii:
Oi! Good catch. Will merge after Travis passes. Thanks for catching this!

nlutsenko:
Awesome catch!

PR #314, opened 2015-09-23
Update Gemfile.

nlutsenko:
Run `bundle update`, remove `slather` as it's not needed anymore.


richardjrossiii:
Gems, gems are truly, truly outrageous.

Issue #179, opened 2015-09-06
watchOS 2 Support


nlutsenko:
This issue is for tracking watchOS 2 support, mainly via Cocoapods.
To use the beta version of Parse SDK in your watchOS 2 app - add the following to your podfile:
```ruby
pod 'Parse', :git => 'git@github.com:ParsePlatform/Parse-SDK-iOS-OSX.git', :branch => 'master'
```

We are aiming to get this very soon out, please comment here if you find any issues using Parse SDK with watchOS 2.

Implementation todo list:
- [x] Add basic target for watchOS 2 SDK (compiler warnings are ok, since we are not going to add this to tests just yet)
- [x] Decouple Reachability from PFEventuallyQueue/PFCommandCache/PFPinningEventuallyQueue (this is going to still be used for Analytics)
- [x] Disable ability to `saveEventually`/`deleteEventually`, since it doesn't quite make sense on the watchOS
- [x] Update PFDevice to use WatchKit APIs on watchOS
- [x] Update PFApplication to be a no-op on watchOS
- [x] Update PFLocationManager to use compatible APIs when running on watchOS
- [x] Disable extensions data sharing APIs (they don't work on watchOS)
- [x] Disable PFPush (at least for the first release)
- [x] Disable PFInstallation (at least for the first release, unless we build PushKit support)
- [x] Disable PFPurchase (no StoreKit support on watchOS)
- [x] Disable PFProduct (doesn't really make sense without PFPurchase)
- [x] Disable PFNetworkActivityIndicatorManager (no indicator on watchOS)
- [x] Add watchOS 2 SDK to Travis-CI just for building
- [x] Test all the things! (no XCTest for watchOS, bummer!)
- [x] Update podspec to include only watchOS specific things
- [x] Add watchOS starter project (optional)
- [x] Add watchOS 2 SDK to deployment packages

Development is happening in master, though we aren't going to ship the watchOS version until this list is finished.
If you want to play with Parse SDK for watchOS - simply compile the Parse-watchOS target available in Parse.xcworkspace.

_This comment is constantly updating, sometimes without you noticing_

bmueller:
Also, could we please get support for PushKit when sending push notifications? PushKit is required to update watchOS 2 complications remotely via push notification.

nlutsenko:
[...]

PR #322, opened 2015-09-24
Exclude PFPush, PFAlertView, PFNetworkIndicatorManager from watchOS target.

nlutsenko:
Constributes to #179


richardjrossiii:
surewhynot

PR #323, opened 2015-09-24
Fixed PFAnalytics for watchOS.

nlutsenko:
Contributes to #179

### DIFF
--- a/Parse/PFAnalytics.m
+++ b/Parse/PFAnalytics.m
@@ -25,9 +25,11 @@
 ///--------------------------------------
 
 + (BFTask *)trackAppOpenedWithLaunchOptions:(NSDictionary *)launchOptions {
-#if TARGET_OS_IPHONE
+#if TARGET_OS_WATCH
+    NSDictionary *userInfo = nil;
+#elif TARGET_OS_IOS
     NSDictionary *userInfo = launchOptions[UIApplicationLaunchOptionsRemoteNotificationKey];
-#else
+#elif PF_TARGET_OS_OSX
     NSDictionary *userInfo = launchOptions[NSApplicationLaunchUserNotificationKey];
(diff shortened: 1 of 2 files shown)


Parse/PFAnalytics.m
```diff
@@ -25,9 +25,11 @@ @implementation PFAnalytics
 ///--------------------------------------
 
 + (BFTask *)trackAppOpenedWithLaunchOptions:(NSDictionary *)launchOptions {
-#if TARGET_OS_IPHONE
+#if TARGET_OS_WATCH
+    NSDictionary *userInfo = nil;
+#elif TARGET_OS_IOS
     NSDictionary *userInfo = launchOptions[UIApplicationLaunchOptionsRemoteNotificationKey];
-#else
+#elif TARGET_OS_MAC
```
richardjrossiii: Seeing as how this is set to true for all iOS derivative platforms (e.g. tvOS as well). I wonder if we can come up with a better macro here...

nlutsenko: I can define PF_TARGET_OS_OSX and combine !TARGET_OS_WATCH, !TARGET_OS_IOS and eventually !TARGET_OS_TV.
What do you think?

richardjrossiii: We can do it that way, sure. I mean, we're still screwed when apple releases a new platform, but I think it's a bit less misleading that way.

nlutsenko: Great, did you have something specific in mind?

richardjrossiii: I wish I did, but I can't think of anything off of the top of my head. Just stick with `PF_TARGET_OS_OSX`, and we can update that macro if we have any brilliant ideas on how to make it better later.

PR #328, opened 2015-09-25
Exclude PFProduct, PFPush internal things from watchOS target.

nlutsenko:
Contributes to #179.

### DIFF
--- a/Parse/Internal/Commands/PFRESTPushCommand.h
+++ b/Parse/Internal/Commands/PFRESTPushCommand.h
@@ -9,11 +9,13 @@
 
 #import "PFRESTCommand.h"
 
+#import <Parse/PFConstants.h>
+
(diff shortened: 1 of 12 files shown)


Parse/Internal/Commands/PFRESTPushCommand.h
```diff
@@ -13,7 +13,7 @@
 
 NS_ASSUME_NONNULL_BEGIN
 
-@interface PFRESTPushCommand : PFRESTCommand
+PF_WATCH_UNAVAILABLE @interface PFRESTPushCommand : PFRESTCommand
```
richardjrossiii: Please do pus related items in a separate PR.

nlutsenko: Renamed and rebased.

PR #355, opened 2015-09-30
Fixed potential null dereference when creating files with nil error.

nlutsenko:
If error pointer is `nil` here - we are going to have a null dereference.


richardjrossiii:
Whoops! Good catch.

PR #356, opened 2015-10-01
Enable clang modules by default.

nlutsenko:
Since we explicitly disable module autolink for frameworks - we are all good even if we enable clang modules. This might potentially improve compilation time.


nlutsenko:
I am amazed! 
If you actually look at the numbers, it's crazy how faster Travis has become, simply due to this change!

~45min for all our builds combined without modules
~36min with modules
Which gets us to ~20% improvement!

richardjrossiii:
Can we do the `@import` things in a more sweeping way in another diff? I don't think we have all of them here, and `#import` style includes are automatically converted to `@import` by clang anyway.

nlutsenko:
Split the last piece into a separate PR.

PR #378, opened 2015-10-07
Mark extension-specific APIs as unavailable for watchOS.

nlutsenko:
This APIs don't quite make sense on watchOS, since there is no ability to use main app + extensions on watchOS.
Contributes to #179.

### DIFF
--- a/Parse/Parse.h
+++ b/Parse/Parse.h
@@ -103,7 +103,7 @@ PF_ASSUME_NONNULL_BEGIN
 
  @param groupIdentifier Application Group Identifier to share data with.
  */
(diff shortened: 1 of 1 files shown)


Bogidon:
Hey @nlutsenko. I spent a lot of time today trying to debug why this method wasn't working on watchOS:
```swift
Parse.enableDataSharingWithApplicationGroupIdentifier(groupIdentifier: NSString, bundleIdentifier: NSString)
```

This GH issue definitely answers that, though it's mentioned nowhere else. [The docs](https://www.parse.com/docs/ios/guide#extensions) sadly seem to be oblivious to the change, still telling users to use this method for WatchKit apps. Should I edit the docs...?

nlutsenko:
@Bogidon, that would be super helpful, thank you.
One thing that would be helpful to note in the docs that they are still valid if targeting watchOS 1 apps.

Bogidon:
@nlutsenko ah ok. Will do in the next few days.

PR #399, opened 2015-10-10
Fix logic for file download error handling.

nlutsenko:
@ghugues thanks for finding this! Tremendous help!


ghugues:
Ah ah no problem, the issue surfaced pretty quickly when I launched my app after asking cocoapods to use master !

ghugues:
Well It seems that doing boolean logic on a Saturday is worse than on a Friday ! (you should check that PR again I think ^^)

nlutsenko:
Should be better now, I hope?

richardjrossiii:
Lol. I'm officially revoking your binary operator card, you now are only allowed to use unary operators.

PR #405, opened 2015-10-12
Fix all new warnings.

nlutsenko:
Split into multiple commits, but will squash together after review.
Depends on #403

### DIFF
--- a/Parse/Internal/Query/Utilities/PFQueryUtilities.m
+++ b/Parse/Internal/Query/Utilities/PFQueryUtilities.m
@@ -96,7 +96,7 @@
                  } comparisonBlock:^NSPredicate *(NSComparisonPredicate *comparison) {
                      NSPredicateOperatorType newType;
                      NSComparisonPredicateModifier newModifier = comparison.comparisonPredicateModifier;
-                     SEL customSelector;
+                     SEL customSelector = NULL;
 
                      switch (comparison.predicateOperatorType) {
                          case NSEqualToPredicateOperatorType: {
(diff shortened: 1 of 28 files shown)


Parse/Internal/Query/Utilities/PFQueryUtilities.m
```diff
@@ -86,17 +86,11 @@ + (NSPredicate *)_negatePredicate:(NSPredicate *)predicate {
...
-                                         format:@"This compound predicate cannot be negated. (%zd)",
-                              compound.compoundPredicateType];
-                             return nil;
-                         }
                      }
                  } comparisonBlock:^NSPredicate *(NSComparisonPredicate *comparison) {
                      NSPredicateOperatorType newType;
                      NSComparisonPredicateModifier newModifier = comparison.comparisonPredicateModifier;
-                     SEL customSelector;
+                     SEL customSelector = nil;
```
richardjrossiii: This needs to be `NULL`, not nil. `SEL` is not an object type.

Parse/Internal/Query/Utilities/PFQueryUtilities.m
```diff
@@ -86,17 +86,11 @@ + (NSPredicate *)_negatePredicate:(NSPredicate *)predicate {
                              }
                              return [NSCompoundPredicate andPredicateWithSubpredicates:newSubpredicates];
                          }
-                         default: {
```
richardjrossiii: I still like having default use cases here. It protects us from some stupid person coming along with a custom predicate and destroying anything. I know it's less optimal, but silent failures are bad.

PR #410, opened 2015-10-14
Add development version of the Podspec with watchOS 2 support.

nlutsenko:
We don't want to ship this in the main podspec just yet, but still be able to test it on Travis and if someone wants to target it from source - it will work beautifully.
Contributes to #179

### DIFF
--- a/Parse.dev.podspec
+++ b/Parse.dev.podspec
@@ -1,0 +1,67 @@
+Pod::Spec.new do |s|
+  s.name             = 'Parse'
+  s.version          = '1.9.0'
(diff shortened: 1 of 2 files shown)


nlutsenko:
Please note that I didn't add it to the tests, since we don't want to ship it just yet, but it gives people an ability to use it through a direct reference in their Podfile.

Biggest reason for not running lint on it - Cocoapods complains about name of the file not matching name of the pod, and if we change the name of the pod - dependencies like ParseUI or FB/Twitter will fail to resolve dependencies.

PR #416, opened 2015-10-15
Unify names for all SDK info.plists.

nlutsenko:
Add watchOS info.plists to the xcodeproj (it was missing), update paths and names.


nlutsenko:
Rebased, and fixed the deployment tests. Now should pass.

PR #429, opened 2015-10-16
Fix a potential deadlock that could occur when encoding for REST.

### LINKED ISSUE #398: Deadlock
I have no idea how and why this is happening but it's happening. Quite frequently now. It happens in different places, in different situations. I don't know an exact sequence of steps to reproduce. Here's a full backtrace: https://gist.github.com/xissburg/a65549d78abbe0a8572f

I am just accessing a PFObject property and the deadlock happens.

richardjrossiii:
We didn't actually need to hold the lock here, and if we did, another thread could have acquired these locks in another order, causing a deadlock.

Should help to fix #398, #11, and a few others.

### DIFF
--- a/Parse/PFObject.m
+++ b/Parse/PFObject.m
@@ -958,25 +958,33 @@ static BOOL PFObjectValueIsKindOfMutableContainerClass(id object) {
  */
 - (NSDictionary *)RESTDictionaryWithObjectEncoder:(PFEncoder *)objectEncoder
                                 operationSetUUIDs:(NSArray **)operationSetUUIDs {
+    NSArray *operationQueue = nil;
+    PFObjectState *state = nil;
+    BOOL deleting = NO;
     @synchronized (lock) {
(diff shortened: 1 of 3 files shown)


Parse/PFObject.m
```diff
@@ -958,25 +958,33 @@ + (BFTask *)_migrateObjectInBackgroundFromFile:(NSString *)fileName
...
-                                               state:state
-                                   operationSetQueue:operationSetQueue];
+        state = self._state;
+        operationQueue = [[NSArray alloc] initWithArray:operationSetQueue copyItems:YES];
+        deleting = _deletingEventually;
     }
+
+    return [self RESTDictionaryWithObjectEncoder:objectEncoder
+                               operationSetUUIDs:operationSetUUIDs
+                                            state:state
```
nlutsenko: Nit: Re-indent.

nlutsenko:
Fix tests, re-indent and ship!

Issue #250, opened 2015-09-14
Parse SDK for tvOS


waterclock:
Please make available a tvOS-compatible version of the Parse frameworks. Apple TV apps will not be able to compile against the iOS frameworks upon official release of Apple's tvOS SDK.

nlutsenko:
Hey @waterclock, absolutely!
We are working hard to make it happen, and tracking this for release soon. Stay tuned.

waterclock:
Great news, thanks!

vikmeup:
+1

WillChilcutt:
+1

artisanglobal:
+1

gabek:
+1

vikmeup:
@nlutsenko any estimates? Thanks.

nlutsenko:
Amazing to see this level of excitement.
Let me output some updates here:
- Since there is no local storage or push notifications on tvOS as of right now, we are investigating what parts of the SDK should be evicted from tvOS target
- For user authentication and analytics tracking we still need a place to put currentUser and installationId (these are just files right now, but you don't really want them to be synced to all your devices)
- We are aiming to nail down the list of all the things that we need to build/change/remove/rinse/repeat to support tvOS by mid next week
- This list should appear on this issue, so we are publicly tracking it and can develop in the wild open (as well as accept contributions)

Any feedback so far about parts that you think are the most important to have, or even something new that you can see us building are very welcome!

WillChilcutt:
As you pointed out, since we cannot store data locally, the first thing to get working is sending and retrieving information. I think the "nice to haves" like the custom UI and analytic, etc. can come later as we have a few weeks before apps will be released, but having the ability to store data remotely is the most important issue right now.

charlync:
+1

reggatta:
Maybe there's a way to put a wrapper around the javascript API or create a lightweight IOS SDK that allows the basics?  It seems that the tvOS limits are such that maybe a better approach is to provide connectivity to Parse cloud code.

waterclock:
Llightweight local storage is available via NSUserDefaults (up to 500K). Also, caches can be written to disk for things like image caching, but they can be deleted by the system so will need to test for their existence prior to reading. 

Definitely need the basic Objective-C API of Parse.framework and ParseUI.framework. All the social authentication frameworks can wait till twitter and Facebook figure out how to offer third-party authentication on a platform without a web view.

[...]

PR #430, opened 2015-10-17
Add basic tvOS SDK target.

nlutsenko:
First step in big game.
Contributes to #250

### DIFF
--- a/Configurations/Bolts-tvOS.xcconfig
+++ b/Configurations/Bolts-tvOS.xcconfig
@@ -1,0 +1,10 @@
+//
+// Copyright (c) 2015-present, Parse, LLC.
+// All rights reserved.
+//
+// This source code is licensed under the BSD-style license found in the
+// LICENSE file in the root directory of this source tree. An additional grant
+// of patent rights can be found in the PATENTS file in the same directory.
+//
+
+SDKROOT = appletvos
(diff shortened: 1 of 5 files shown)


waterclock:
Does this compile? I kinda need a tvOS Parse framework that will compile, even if it's wildly broken as I need to submit a provisional build to iTunesConnect to test some StoreKit code. Thanks!

nlutsenko:
Nope, not yet... :)
Will make it compile very very soon.

Configurations/Bolts-tvOS.xcconfig
```diff
@@ -0,0 +1,10 @@
+//
+// Copyright (c) 2015-present, Parse, LLC.
+// All rights reserved.
+//
+// This source code is licensed under the BSD-style license found in the
+// LICENSE file in the root directory of this source tree. An additional grant
+// of patent rights can be found in the PATENTS file in the same directory.
+//
+
+SDKROOT = appletvos
```
richardjrossiii: I don't think we want this file.

fvimagination:
Hi, just wondering if you will post something here when Parse will be ok with tvOS.
Thanks!

eytanschulman:
I want to submit my tv app today or tomorrow, and Parse Analytics would be killer - i'd really really like some analytics in my app. Does this work yet?

PR #431, opened 2015-10-18
spelling

hhanesand:


nlutsenko:
Nice catch! Thanks!
Will merge in straight away after Travis finishes building, since we have required checks enabled.

PR #436, opened 2015-10-19
Remove PFPush*/PFAlertView from tvOS target.

nlutsenko:
Contributes to #250.


richardjrossiii:
We should open a task to eventually add PFPush back in for tvOS, as it would appear that tvOS will support push notifications.

nlutsenko:
We are tracking this in general, right?
It's not 100% confirmed, yes they opened the API, but it doesn't yet work.
Let me open an issue... Maybe this is finally coming today, since it's down for maintenance right now?

PR #439, opened 2015-10-20
Remove potential deadlock from encoding PFUser.

nlutsenko:
No need for user.lock here, since we removed the super class one as well and we are guaranteed to be passed in a copy of opSetQueue.
Has a potential to resolve yet another deadlock specifically when using users.

### DIFF
--- a/Parse/PFUser.m
+++ b/Parse/PFUser.m
@@ -652,23 +652,21 @@ static BOOL revocableSessionEnabled_;
                                             state:(PFObjectState *)state
                                 operationSetQueue:(NSArray *)queue
                           deletingEventuallyCount:(NSUInteger)deletingEventuallyCount {
-    @synchronized (self.lock) {
-        NSMutableArray *cleanQueue = [queue mutableCopy];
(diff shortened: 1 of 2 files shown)


Parse/PFUser.m
```diff
@@ -652,23 +652,21 @@ - (NSDictionary *)RESTDictionaryWithObjectEncoder:(PFEncoder *)objectEncoder
...
-                                    operationSetUUIDs:operationSetUUIDs
-                                                state:state
-                                    operationSetQueue:cleanQueue
-                                   deletingEventually:isDeletingEventually];
-    }
+    NSMutableArray *cleanQueue = [queue mutableCopy];
+    [queue enumerateObjectsUsingBlock:^(PFOperationSet *operationSet, NSUInteger idx, BOOL *stop) {
+        // Remove operations for `password` field, to not let it persist to LDS.
+        if (operationSet[PFUserPasswordRESTKey]) {
+            operationSet = [operationSet copy];
```
richardjrossiii: Because we have an operation set copy, we shouldn't need to copy it again here.

nlutsenko: I feel like it's an implementation detail of the superclass method, so probably safer to leave it here.

PR #456, opened 2015-10-22
Embed bitcode for all iOS 9.* SDKs.

nlutsenko:


Vendor/xctoolchain
```diff
@@ -1 +1 @@
-Subproject commit 65c31f2f70fe3c0c78b19e19f23b8c0fbcfd6876
+Subproject commit 94b6a06c0c649d71a11a4d2f7ace28bc8c865acb
```
richardjrossiii: This shouldn't be here..

PR #457, opened 2015-10-22
Updated README badge styles.

nlutsenko:
cc @andrewimm


andrewimm:
The badges fit so perfectly now. This will cause them to overflow into two lines.

screen shot 2015-10-22 at 4 48 34 pm

nlutsenko:
I explicitly made it be on the next line by adding a line break.
Also, I am going to change the codecov badge slightly right now.

PR #458, opened 2015-10-23
Update Bolts to 1.4.0.

nlutsenko:
🔩🔩🔩!


nlutsenko:
Please note - we need recursive submodules, since Bolts now depends on xctoolchain and Travis-CI doesn't checkout submodules by default (that's why I disabled the automatic Travis one).

Issue #466, opened 2015-10-25
Cannot Build In Xcode 7.1


shushutochako:
I upgraded to XCode 7.1 yesterday.
When using Cocoapods,  the module will no longer build for me with the error 
```
include of non-modular header inside framework module 'Parse.Parse'
#import <Parse/PFACL.h>

include of non-modular header inside framework module 'Parse.Parse'
#import <Parse/PFAnalytics.h>
・
・
more
```
Is there a workaround?

nlutsenko:
Sounds super strange, since we have tests passing on every commit running Xcode 7.1, which explicitly test CocoaPods linter. I would suggest tyin to clean your derived data and/or reinstalling all your pods.

shushutochako:
Thank you for answering. I tried reinstalling and clean but it was the same result.
The image below is the capture of the error.

![capture](https://cloud.githubusercontent.com/assets/8078429/10715293/d6c04d8c-7b54-11e5-9f3f-729c2de94e5d.png)

Incidentally version of cocoapod is 0.39.0 and Parse is 1.9.1.

BellAppLab:
Same thing here...

shushutochako:
Hi nlutsenko. 
This issue has been resolved by add description of the following to Podfile.
```
post_install do |installer|
    `find Pods -regex 'Pods/Parse.*\\.h' -print0 | xargs -0 sed -i '' 's/\\(<\\)Parse\\/\\(.*\\)\\(>\\)/\\"\\2\\"/'`
end
```
I'm sorry for the inconvenience.

fluidsonic:
@shushutochako adding code to the Podfile is a workaround, not a solution.

Same problem here with CocoaPods 0.39.0, Xcode 7.1 & Parse 1.9.1.

@nlutsenko why is the style `#import <Parse/XYZ.h>` used instead of the typical `#import "XYZ.h"`?
Changing all imports fixes the problem like @shushutochako's workaround does.

BellAppLab:
One million times THIS. 

Same goes for Bolts. Also, referencing Bolts with `@import Bolts;` rather than `#import <Bolts/BFTask.h>`.

I've been doing the refactoring here, @nlutsenko and will be glad to push them through.

fluidsonic:
Looks like Bolts imports are also affected.
Fixing just the `#import <Parse/…>` leads to the next error:

```
[...]

PR #469, opened 2015-10-26
Fixed #imports to properly support Xcode 7.1

fluidsonic:
A framework's header files can no longer import other own header files using angular import syntax (`#import <Parse/…>`) as of Xcode 7.1. That leads to a lot errors when importing (not compiling) this framework.

Example:
```
include of non-modular header inside framework module 'Parse.Parse'
#import <Parse/PFACL.h>
```

Properly using `#import "XYZ.h"` instead of `#import <Parse/XYZ.h>` fixes this issue.

Additionally Parse imports some specific headers from the Bolts framework directly which also violates semantic import. Instead of `#import <Bolts/XYZ.h>` it must import Bolts framework as a whole.

Since `@import` is not available for apps targeting iOS 7 or lower and requires CocoaPods users to enable frameworks (using `use_frameworks!`), `Parse+Bolts.h` was added as a workaround. It imports `<Bolts/Bolts-umbrella.h>` if it exists (it is generated by CocoaPods) or else Bolts' default umbrella header `<Bolts/Bolts.h>`.

Also note that this fix requires the same issue to be fixed in Bolts. See https://github.com/BoltsFramework/Bolts-iOS/pull/179

Solves #466


fluidsonic:
Although this pull request fixes the issue this is a widespread problem which is being tracked and most likely fixed by CococaPods itself: https://github.com/CocoaPods/CocoaPods/issues/4420

PR #484, opened 2015-10-27
Replace all usage of nullability annotations via macros with generic one.

nlutsenko:
This will enforce usage of Xcode 6.3+, but is very much intentional.


Parse/PFObject.h
```diff
@@ -39,7 +39,7 @@ NS_REQUIRES_PROPERTY_DEFINITIONS
     // Every time a save is started, a new dictionary is added to the end.
     // Whenever a save completes, the new data is put into fetchedData, and
     // a dictionary is removed from the start.
-    NSMutableArray *PF_NULLABLE_S operationSetQueue;
+    NSMutableArray *__nullable operationSetQueue;
```
richardjrossiii: We really need to get rid of this sometime soon.

PR #503, opened 2015-11-03
Return nil if sysctlbyname call fails

Chilledheart:


nlutsenko:
@Chilledheart This looks good. Any chance you can outline an issue that you are running into here?
Or this is purely to safeguard the code?

`stringWithUTF8String:` initialization from an empty string is still valid and won't crash, but I think you are right that this approach is better...

cc @richardjrossiii

Parse/Internal/PFDevice.m
```diff
@@ -27,14 +27,19 @@
     const char *charName = [name UTF8String];
 
     size_t size;
-    sysctlbyname(charName, NULL, &size, NULL, 0);
+    if (sysctlbyname(charName, NULL, &size, NULL, 0) < 0) {
```
richardjrossiii: This should be `== 0`, not `< 0`. Theoretically an implementation could return `1` to indicate error as well.

Parse/Internal/PFDevice.m
```diff
@@ -27,14 +27,19 @@
...
     if (answer == NULL) {
         return nil;
     }
 
-    sysctlbyname(charName, answer, &size, NULL, 0);
+    if (sysctlbyname(charName, answer, &size, NULL, 0) < 0) {
+      free(answer);
+      return nil;
+    }
     NSString *string = [NSString stringWithUTF8String:answer];
```
richardjrossiii: If we're truly being safe here, we should manually pass the length of the bytes here. There's nothing in the `sysctl` man page that indicates that the strings returned through it should be NUL-terminated, even if they are in practice.

richardjrossiii:
There's also the chance for a TOCTTOU bug here, though not sure how we could work around it. One more reason to explicitly pass length to the `NSString` constructor.

Parse/Internal/PFDevice.m
```diff
@@ -27,14 +27,19 @@
...
+    }
     char *answer = (char*)malloc(size);
 
     if (answer == NULL) {
         return nil;
     }
 
-    sysctlbyname(charName, answer, &size, NULL, 0);
+    if (sysctlbyname(charName, answer, &size, NULL, 0) < 0) {
+      free(answer);
```
richardjrossiii: I also don't like the idea of having free in multiple places... I have a couple of thoughts on ways to clean this up...

1. Use a C99 VLA for answer instead of `malloc`.
2. Refactor to use a single exit point.

This situation right here is why most maintainers of C code are adamant about single exit points, as cleanup only happens in a single place :)

Chilledheart:
@nlutsenko In fact the point is that, if `sysctlbyname` fails, `answer` is not guaranteed to be a null string since it is not initialized and therefore `string` might be a random string.

Parse/Internal/PFDevice.m
Chilledheart: I think not, because sysctlbyname always returns `-1` on failure or `0` on success (as `man 3 sysctl` says)

Parse/Internal/PFDevice.m
Chilledheart: That's cool.

Parse/Internal/PFDevice.m
Chilledheart: Maybe we can allocate memory from stack instead of heap here?

Parse/Internal/PFDevice.m
Chilledheart: Hmm, it seems Foundation framework has deprecated the such constructor which requires both pointer and length.

Parse/Internal/PFDevice.m
richardjrossiii: For OSX, that's correct. Not mandated by the POSIX standard IIRC, however.

Parse/Internal/PFDevice.m
richardjrossiii: `initWithBytes:length:encoding:` should be the proper method.

Parse/Internal/PFDevice.m
richardjrossiii: Only worry there is a bad value being set by another process that exceeds the stack size and smashes it....

I suppose in that case we can claim "undefined behavior" and be done with it.

Chilledheart: That's fair.

Parse/Internal/PFDevice.m
Chilledheart: > For OSX, that's correct. Not mandated by the POSIX standard IIRC, however.
Okay. I can understand it.
> This should be == 0, not < 0. Theoretically an implementation could return 1 to indicate error as well.
Did you mean '!= 0' here ?

nlutsenko:
@Chilledheart yup, misread the variables. You are right - using malloc there is bad. Random string all the way.

Parse/Internal/PFDevice.m
richardjrossiii: Yes, my bad.

PR #492, opened 2015-10-29
Reimplement file persistence via PFPersistenceController + PFPersistenceGroup, migrate applicationId storage to it.

nlutsenko:
- PFPersistenceController
  - PFFilePersistenceGroup (iOS/OSX/watchOS)
  - PFUserDefaultsPersistenceGroup (tvOS)
  
Also fully migrated our applicationId storage to be based purely on PFPersistenceController with validation and merrymaking.
Contributes to #250.

PR #507, opened 2015-11-03
Move PFConfig to use persistence groups.

nlutsenko:
This moves all PFConfig related code to use persistence groups instead of file manager APIs.
Including PFCurrentConfigController, PFConfigController.

Depends on #492.


nlutsenko:
Tests are going to fail on this, since I didn't update them yet to mock persistence group, but please take a look at the actual code.

richardjrossiii:
Seems logical to me. Fix tests and ship.

PR #509, opened 2015-11-04
Make PFDeviceSysctlByName more safe and easy to maintain

Chilledheart:
Changes made according to the comments in #503

### DIFF
--- a/Parse/Internal/PFDevice.m
+++ b/Parse/Internal/PFDevice.m
@@ -25,24 +25,27 @@
 
 static NSString *PFDeviceSysctlByName(NSString *name) {
     const char *charName = [name UTF8String];
+    NSString *string = nil;
+    size_t size = 0;
+    char *answer = NULL;
 
-    size_t size;
-    if (sysctlbyname(charName, NULL, &size, NULL, 0) < 0) {
-      return nil;
-    }
-    char *answer = (char*)malloc(size);
+    do {
+        if (sysctlbyname(charName, NULL, &size, NULL, 0) != 0) {
+            break;
+        }
+        answer = (char*)malloc(size);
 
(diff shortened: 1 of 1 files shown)


nlutsenko:
Looks better, love that we are using proper encoding, not a big fan of goto.
Deferring to @richardjrossiii.

richardjrossiii:
Getting much closer. Let's see if we can avoid using `goto` here, as it has one major con with our usage: Undefined behavior with ARC enabled.

If ARC is enabled and you use a `goto` statement with an ARC-managed variable in scope, it's undefined whether or not ARC will properly clean-up that variable. In this particular case, we never have a variable who's scope gets improperly exited, but I'm still not a fan of using `goto` with ARC because of the maintenance issues involved.

Can we change this into a `do { } while(0);` loop, and just `break` in error cases? Remember that its well-defined behavior to call `free(NULL)`, so I think it will create much cleaner code. In the end.

Parse/Internal/PFDevice.m
```diff
@@ -25,24 +25,28 @@
 
 static NSString *PFDeviceSysctlByName(NSString *name) {
     const char *charName = [name UTF8String];
-
+    NSString *string = nil;
     size_t size;
```
richardjrossiii: For consistency, let's give this an initial value too.

Chilledheart:
@richardjrossiii It is, definitely, better to use `do-while` instead of `goto` in the mixed objective-c code.

I have updated the code. Please double-check the code before we can merge this.

richardjrossiii:
Love it! @nlutsenko, when you have a chance to look it over, merge me up, scotty!

nlutsenko:
Aye, captain, energizing...

Issue #290, opened 2015-09-18
Crash within Parse framework when it internally uses PFSQLiteDatabase


rdbayer:
On Parse 1.8.3, we are periodically experiencing a crash.  The relevant call appears to be [PFSQLiteDatabase _executeQueryAsync:withArgumentsInArray:cachingEnabled:]  called by [BFTask continueWithExecutor:block:cancellationToken:]_block_invoke_2.  Will paste the full stack trace for all threads as a comment.

It looks like this is happening as the result of a PFQuery.findObjectInBackgroundWithBlock() call

nlutsenko:
Hey @rdbayer, thanks for the report!
Stack trace would be amazing, since without it it's going to be very hard to track it down.
Also, getting a full blown stack trace would be much more preferred to the regular Xcode one.
(aka running `thread backtrace all` from lldb console)

rdbayer:
Full stack trace for all threads:

===

Thread : com.apple.main-thread
0  libsystem_kernel.dylib         0x0000000198380c84 semaphore_timedwait_trap + 8
1  libdispatch.dylib              0x000000019825eaa4 _dispatch_group_wait_slow + 176
2  libdispatch.dylib              0x0000000198256dcc dispatch_block_wait + 244
3  BaseBoard                      0x000000018a2b536c -[NSObject(BaseBoard) bs_performSynchronously:timeout:] + 160
4  FrontBoardServices             0x000000018cb82178 -[FBSUIApplicationSystemService _getBadgeValue] + 184
5  FrontBoardServices             0x000000018cb823c0 -[FBSUIApplicationSystemService badgeNumber] + 24
6  Breakaway                      0x00000001003343e0 -[PFApplication iconBadgeNumber] + 558360
7  Breakaway                      0x000000010033441c -[PFApplication setIconBadgeNumber:] + 558420
8  Breakaway                      0x000000010032ef14 -[PFInstallation setObject:forKey:] + 536652
9  Breakaway                      0x00000001002df31c -[PFObject setObject:forKeyedSubscript:] + 210004
10 Breakaway                      0x0000000100332e74 -[PFObjectSubclassingController _forwardSetterInvocation:forProperty:withObject:] + 552876
11 Breakaway                      0x00000001003326d0 -[PFObjectSubclassingController forwardObjectInvocation:withObject:] + 550920
12 Breakaway                      0x00000001002e12a0 -[PFObject forwardInvocation:] + 218072
[...]

PR #511, opened 2015-11-05
Made all accesses to PFSQLiteDatabaseResult & Statement thread-safe.

richardjrossiii:
I think this is what was causing the crashes seen in #290, but I'm not 100% sure.

At the very least, accessing them this way was in violation of the SQLite spec.

### DIFF
--- a/Parse/Internal/LocalDataStore/SQLite/PFSQLiteDatabaseResult.m
+++ b/Parse/Internal/LocalDataStore/SQLite/PFSQLiteDatabaseResult.m
@@ -12,11 +12,13 @@
 #import <sqlite3.h>
 
 #import "PFSQLiteStatement.h"
+#import "PFThreadsafety.h"
 
 @interface PFSQLiteDatabaseResult ()
 
 @property (nonatomic, copy, readonly) NSDictionary *columnNameToIndexMap;
 @property (nonatomic, strong, readonly) PFSQLiteStatement *statement;
+@property (nonatomic, strong, readonly) dispatch_queue_t databaseQueue;
 
 @end
 
@@ -24,9 +26,10 @@
 
 @synthesize columnNameToIndexMap = _columnNameToIndexMap;
 
-- (instancetype)initWithStatement:(PFSQLiteStatement *)stmt {
+- (instancetype)initWithStatement:(PFSQLiteStatement *)stmt queue:(dispatch_queue_t)queue {
     if ((self = [super init])) {
         _statement = stmt;
+        _databaseQueue = queue;
     }
     return self;
 }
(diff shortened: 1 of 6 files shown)


Parse/Internal/LocalDataStore/SQLite/PFSQLiteDatabaseResult.m
```diff
@@ -12,11 +12,20 @@
 #import <sqlite3.h>
 
 #import "PFSQLiteStatement.h"
+#import "PFThreadsafety.h"
+
+// PFThreadsafetySafeDispatchSync, but with a return type.
+#define thread_safe_sync(queue, block) ({                          \
```
nlutsenko: Any chance we can make this a function?
This is not going to be nice (probably), but might feel better than having 2 defines that do the same thing...

Parse/Internal/LocalDataStore/SQLite/PFSQLiteStatement.h
```diff
@@ -19,8 +19,9 @@ typedef struct sqlite3_stmt sqlite3_stmt;
 @interface PFSQLiteStatement : NSObject
 
 @property (atomic, assign, readonly) sqlite3_stmt *sqliteStatement;
+@property (atomic, strong, readonly) dispatch_queue_t databaseQueue;
```
nlutsenko: Do we really need this to be `atomic`? it's readonly and is set on initialization... :)

Parse/Internal/LocalDataStore/SQLite/PFSQLiteStatement.m
```diff
@@ -11,13 +11,23 @@
...
+    PFThreadsafetySafeDispatchSync(queue, ^{ result = block(); }); \
+    result;                                                        \
+})
+
 @implementation PFSQLiteStatement
 
-- (instancetype)initWithStatement:(sqlite3_stmt *)stmt {
+- (instancetype)initWithStatement:(sqlite3_stmt *)stmt queue:(dispatch_queue_t)databaseQueue {
     self = [super init];
     if (!stmt || !self) return nil;
```
nlutsenko: We really need to clean this up at some point... `if (!stmt)` sounds pretty damn bad...

nlutsenko:
Few nits, otherwise lgtm.
Love the idea!

Parse/Internal/LocalDataStore/SQLite/PFSQLiteDatabaseResult.m
richardjrossiii: I'm not sure how we can, seeing as the point of this macro is to be able to easily return any type, not just `id` or `int`.

We could stick it in `PFThreadsafety.h`, perhaps?

nlutsenko: Sounds good, let's do it! :)

richardjrossiii:
PR updated.

nlutsenko:
Awesome! Please ship!

coolnalu:
Awesome! It's HUUUUUUUUGE!

richardjrossiii:
@coolnalu No problem <3

PR #512, opened 2015-11-05
Add API to get user list from permissions in ACL.

soesin:
My project needed to know the list of PFUsers those can read/write the PFObject.


richardjrossiii:
Thanks for the pull request!

I do think that this API is useful, although I think returning a list of PFUser objects is more useful than a list of object IDs.

We'll need to discuss with the rest of the team on their thoughts, as this would need to be implemented in each of our SDKs.

I think the implementation can also be made much more performant as well.

richardjrossiii:
cc @grantland @nlutsenko @hallucinogen @stanleyw Thoughts?

soesin:
Hi @richardjrossiii thank you for the response.

I agree performance of the current implementation is not that good for the readability and I will commit improved one later.

nlutsenko:
There are a lot of nits in the code and block enumeration on dictionaries is actually faster...

Let's have an broad discussion first, though.

Overall, I like the idea! The API naming should be more precise though :)
Additionally, I think it would be nice to split the method into read and write accesses, since they are somewhat different and actually give more possibilities for API usage.
Something like `readAccessUserObjectIds` and `writeAccessUserObjectIds`. The only names that pop in my head sound ugly, so it's tough to come up with a good name. The important aspects here is to explicitly mention that the array being returned has objectIds not users (even though we have generic types).

@grantland @hallucinogen Thoughts?

hhanesand:
Since @richardjrossiii suggested that it would be better if the methods returned `PFUsers` instead, what about `usersWithReadAuthorization` and `usersWithWriteAuthorization`?

richardjrossiii:
@hhanesand Personally, I like that naming much better, and I think it translates much better as a cross-platform API.

PR #513, opened 2015-11-05
Add thread safety tests for InstallationIdentifierStore.

nlutsenko:
Porting a private test. This could safely live here.


richardjrossiii:
surewhynot

PR #514, opened 2015-11-05
Remove any migration of data from old deprecated location.

nlutsenko:
This migration was first implemented in version 0.4.38 of the iOS SDK, according to the version notes and was released around Feb 2012.
We are going to remove this code, so any application upgrading from a version < 0.4.38 will not migrate the data any more.

Since 0.4.38 and 1.10.0 (next version) has at least one major version difference, I feel we should make this call.
Please note that this is going to improve the performance of reading anything from disk, as we are not going to hit the lookup of deprecated path anymore.
Plus, it's going to get so much easier to maintain, as we move to Persistence Groups (disk/UserDefaults for tvOS).

cc @grantland @stanleyw @hallucinogen

### DIFF
--- a/Parse/Internal/PFFileManager.h
+++ b/Parse/Internal/PFFileManager.h
@@ -63,7 +63,6 @@ typedef NS_OPTIONS(uint8_t, PFFileManagerOptions) {
  */
 - (NSString *)parseDefaultDataDirectoryPath;
 - (NSString *)parseLocalSandboxDataDirectoryPath;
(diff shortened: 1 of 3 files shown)


grantland:
This sounds fine to me. Worst case scenario from an upgrade from < 0.4.38 upgrading to > 1.10.0 is that the application perceives it to be a fresh install and the user logs in again and developers upgrading from 0.x.x --> 1.x.x can expect some kind of breaking changes.

richardjrossiii:
BURN IT

PR #552, opened 2015-11-12
Use ObjC literals for Dictionaries/Arrays.

nlutsenko:


richardjrossiii:
LGTM. Only question is does this mean we no longer have a test directly for `-objectForKey:`?

nlutsenko:
We do in places where we need it.
I omitted few of these (ugh, tiresome process) to make sure we test both objectForKey and sub scripting.

PR #553, opened 2015-11-12
Create in-memory eventually queue for analytics events on tvOS.

nlutsenko:
- Created a new class that implements PFEventuallyQueue
- Use in-memory store only for all the commands
- Use the class if compiling for tvOS

Contributes to #250.

### DIFF
--- a/Parse/Internal/PFMemoryEventuallyQueue.h
+++ b/Parse/Internal/PFMemoryEventuallyQueue.h
@@ -1,0 +1,20 @@
+/**
+ * Copyright (c) 2015-present, Parse, LLC.
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree. An additional grant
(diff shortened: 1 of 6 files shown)


Parse/Internal/PFMemoryEventuallyQueue.h
```diff
@@ -0,0 +1,19 @@
...
+ * LICENSE file in the root directory of this source tree. An additional grant
+ * of patent rights can be found in the PATENTS file in the same directory.
+ */
+
+#import "PFEventuallyQueue.h"
+
+PF_IOS_UNAVAILABLE_WARNING
+PF_OSX_UNAVAILABLE_WARNING
+
+PF_IOS_UNAVAILABLE PF_OSX_UNAVAILABLE @interface PFMemoryEventuallyQueue : PFEventuallyQueue
```
richardjrossiii: Shouldn't this be WATCH_UNAVAILABLE as well?

nlutsenko: Duh. Right.

PR #558, opened 2015-11-13
Fix null-termination of system call on getting the full os version.

nlutsenko:
The string that we get back from `sysctlbyname` is usually null-terminated, but we were not handling that case.
This will handle both cases.


richardjrossiii:
Crazy. Crazy stuff.

PR #570, opened 2015-11-16
Add better, more flexible APIs for SDK initialization.

richardjrossiii:
## This PR includes public API Changes

Old initialization looked somewhat outdated, our new APIs allow a usage similar to this:

### ObjC

```objc
[Parse initializeWithConfiguration:[ParseClientConfiguration configurationWithBlock:^(ParseClientConfiguration *configuration) {
    configuration.applicationId = ...;
    configuration.clientKey = ...;
    configuration.localDatastoreEnabled = ... ;

    // Data sharing config, etc.
}];
```

### Swift

```swift
Parse.initializeWithConfiguration(ParseClientConfiguration() {
    $0.applicationId = ""
    $0.clientKey = ""
    $0.localDatastoreEnabled = true
    // etc.
})
````

We believe that this provides a much cleaner interface for initialization. Please feel free to comment below on any feedback that you may have on this

------

Our current SDK initialization is lackluster in that it's split across multiple classes, files, and method calls.

This PR hopes to unify some of these initialization settings int o a single 'configuration' class that can be easily used to intialize parse in a much cleaner way, especially as we add even more initialization options.

### DIFF
--- a/Parse/Internal/ParseClientConfiguration_Private.h
+++ b/Parse/Internal/ParseClientConfiguration_Private.h
@@ -1,0 +1,38 @@
+/**
+ * Copyright (c) 2015-present, Parse, LLC.
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree. An additional grant
+ * of patent rights can be found in the PATENTS file in the same directory.
+ */
+
+#import "ParseClientConfiguration.h"
+
+NS_ASSUME_NONNULL_BEGIN
+
+@interface ParseClientConfiguration ()
+
+@property (nullable, nonatomic, copy, readwrite) NSString *applicationId;
+@property (nullable, nonatomic, copy, readwrite) NSString *clientKey;
+
+@property (nonatomic, assign, readwrite, getter=isLocalDatastoreEnabled) BOOL localDatastoreEnabled;
+
+@property (nullable, nonatomic, copy, readwrite) NSString *applicationGroupIdentifier;
+@property (nullable, nonatomic, copy, readwrite) NSString *containingApplicationBundleIdentifier;
+
+@property (nonatomic, assign, readwrite) NSUInteger networkRetryAttempts;
+
++ (instancetype)emptyConfiguration;
+- (instancetype)initEmpty NS_DESIGNATED_INITIALIZER;
+
+- (void)_resetDataSharingIdentifiers;
+
+@end
+
+// We must implement the protocol here otherwise clang issues warnings about non-matching property declarations.
+// For some reason if the property declarations are on a separate category, it doesn't care.
+@interface ParseClientConfiguration (Private) <ParseMutableClientConfiguration>
+@end
+
+NS_ASSUME_NONNULL_END
--- a/Parse/Internal/ParseManager.m
+++ b/Parse/Internal/ParseManager.m
@@ -90,7 +90,7 @@ static NSString *const _ParseApplicationIdFileName = @"applicationId";
     PFNotDesignatedInitializer();
 }
 
-- (instancetype)initWithApplicationId:(NSString *)applicationId clientKey:(NSString *)clientKey {
+- (instancetype)initWithConfiguration:(ParseClientConfiguration *)configuration {
     self = [super init];
     if (!self) return nil;
 
@@ -108,26 +108,20 @@ static NSString *const _ParseApplicationIdFileName = @"applicationId";
     _controllerAccessQueue = dispatch_queue_create("com.parse.controller.access", DISPATCH_QUEUE_SERIAL);
     _preloadQueue = dispatch_queue_create("com.parse.preload", DISPATCH_QUEUE_SERIAL);
 
-    _applicationId = [applicationId copy];
-    _clientKey = [clientKey copy];
+    _configuration = [configuration copy];
 
     return self;
 }
 
-- (void)configureWithApplicationGroupIdentifier:(NSString *)applicationGroupIdentifier
-                containingApplicationIdentifier:(NSString *)containingApplicationIdentifier
-                          enabledLocalDataStore:(BOOL)localDataStoreEnabled {
-    _applicationGroupIdentifier = [applicationGroupIdentifier copy];
-    _containingApplicationIdentifier = [containingApplicationIdentifier copy];
-
+- (void)startManaging {
     // Migrate any data if it's required.
     [self _migrateSandboxDataToApplicationGroupContainerIfNeeded];
 
     // TODO: (nlutsenko) Make it not terrible!
     [[self.persistenceController getPersistenceGroupAsync] waitForResult:nil withMainThreadWarning:NO];
 
-    if (localDataStoreEnabled) {
-        PFOfflineStoreOptions options = (_applicationGroupIdentifier ?
+    if (self.configuration.localDatastoreEnabled) {
+        PFOfflineStoreOptions options = (self.configuration.applicationGroupIdentifier ?
                                          PFOfflineStoreOptionAlwaysFetchFromSQLite : 0);
         [self loadOfflineStoreWithOptions:options];
     }
@@ -138,8 +132,8 @@ static NSString *const _ParseApplicationIdFileName = @"applicationId";
 ///--------------------------------------
 
 - (void)loadOfflineStoreWithOptions:(PFOfflineStoreOptions)options {
-    PFConsistencyAssert(!_offlineStore, @"Can't load offline store more than once.");
     dispatch_barrier_sync(_offlineStoreAccessQueue, ^{
+        PFConsistencyAssert(!_offlineStore, @"Can't load offline store more than once.");
         _offlineStore = [[PFOfflineStore alloc] initWithFileManager:self.fileManager options:options];
     });
 }
@@ -159,7 +153,7 @@ static NSString *const _ParseApplicationIdFileName = @"applicationId";
 }
 
 - (BOOL)isOfflineStoreLoaded {
-    return (self.offlineStore != nil);
+    return self.configuration.localDatastoreEnabled;
 }
 
 ///--------------------------------------
@@ -226,7 +220,7 @@ static NSString *const _ParseApplicationIdFileName = @"applicationId";
     __block PFKeychainStore *store = nil;
     dispatch_sync(_keychainStoreAccessQueue, ^{
         if (!_keychainStore) {
-            NSString *bundleIdentifier = (_containingApplicationIdentifier ?: [[NSBundle mainBundle] bundleIdentifier]);
+            NSString *bundleIdentifier = (self.configuration.containingApplicationBundleIdentifier ?: [[NSBundle mainBundle] bundleIdentifier]);
             NSString *service = [NSString stringWithFormat:@"%@.%@", bundleIdentifier, PFKeychainStoreDefaultService];
             _keychainStore = [[PFKeychainStore alloc] initWithService:service];
         }
@@ -241,8 +235,8 @@ static NSString *const _ParseApplicationIdFileName = @"applicationId";
     __block PFFileManager *fileManager = nil;
     dispatch_sync(_fileManagerAccessQueue, ^{
         if (!_fileManager) {
-            _fileManager = [[PFFileManager alloc] initWithApplicationIdentifier:self.applicationId
-                                                     applicationGroupIdentifier:self.applicationGroupIdentifier];
+            _fileManager = [[PFFileManager alloc] initWithApplicationIdentifier:self.configuration.applicationId
+                                                     applicationGroupIdentifier:self.configuration.applicationGroupIdentifier];
         }
         fileManager = _fileManager;
     });
@@ -278,7 +272,7 @@ static NSString *const _ParseApplicationIdFileName = @"applicationId";
             return [[NSString alloc] initWithData:data encoding:NSUTF8StringEncoding];
         }] continueWithSuccessBlock:^id(BFTask *task) {
             if (task.result) {
-                if ([self.applicationId isEqualToString:task.result]) {
+                if ([self.configuration.applicationId isEqualToString:task.result]) {
                     // Everything is valid, no need to remove, set applicationId here.
                     return nil;
                 }
@@ -287,15 +281,15 @@ static NSString *const _ParseApplicationIdFileName = @"applicationId";
                 [self.keyValueCache removeAllObjects];
             }
             return [[group removeAllDataAsync] continueWithSuccessBlock:^id(BFTask *task) {
-                NSData *applicationIdData = [self.applicationId dataUsingEncoding:NSUTF8StringEncoding];
+                NSData *applicationIdData = [self.configuration.applicationId dataUsingEncoding:NSUTF8StringEncoding];
                 return [group setDataAsync:applicationIdData forKey:_ParseApplicationIdFileName];
             }];
         }] continueWithBlock:^id(BFTask *task) {
             return [group endLockedContentAccessAsyncToDataForKey:_ParseApplicationIdFileName];
         }];
     };
-    return [[PFPersistenceController alloc] initWithApplicationIdentifier:self.applicationId
-                                               applicationGroupIdentifier:self.applicationGroupIdentifier
+    return [[PFPersistenceController alloc] initWithApplicationIdentifier:self.configuration.applicationId
+                                               applicationGroupIdentifier:self.configuration.applicationGroupIdentifier
                                                    groupValidationHandler:validationHandler];
 }
 
@@ -319,8 +313,9 @@ static NSString *const _ParseApplicationIdFileName = @"applicationId";
     dispatch_sync(_commandRunnerAccessQueue, ^{
         if (!_commandRunner) {
             _commandRunner = [PFURLSessionCommandRunner commandRunnerWithDataSource:self
-                                                                      applicationId:self.applicationId
-                                                                          clientKey:self.clientKey];
+                                                                      retryAttempts:self.configuration.networkRetryAttempts
+                                                                      applicationId:self.configuration.applicationId
+                                                                          clientKey:self.configuration.clientKey];
         }
         runner = _commandRunner;
     });
@@ -455,7 +450,7 @@ static NSString *const _ParseApplicationIdFileName = @"applicationId";
     // There is no need to migrate anything on OSX, since we are using globally available folder.
 #if TARGET_OS_IOS
     // Do nothing if there is no application group container or containing application is specified.
-    if (!self.applicationGroupIdentifier || self.containingApplicationIdentifier) {
+    if (!self.configuration.applicationGroupIdentifier || self.configuration.containingApplicationBundleIdentifier) {
         return;
     }
 
--- a/Parse/Parse.h
+++ b/Parse/Parse.h
@@ -9,6 +9,7 @@
 
 #import <Foundation/Foundation.h>
 
+#import <Parse/ParseClientConfiguration.h>
 #import <Parse/PFACL.h>
 #import <Parse/PFAnalytics.h>
 #import <Parse/PFAnonymousUtils.h>
@@ -65,6 +66,22 @@ NS_ASSUME_NONNULL_BEGIN
  @param clientKey The client key of your Parse application.
  */
 + (void)setApplicationId:(NSString *)applicationId clientKey:(NSString *)clientKey;
+
+/**
+ Sets the configuration to be used for the Parse SDK.
+
+ @note Re-setting the configuration after having previously sent requests through the SDK results in undefined behavior.
+
+ @param configuration The new configuration to set for the SDK.
+ */
++ (void)initializeWithConfiguration:(ParseClientConfiguration *)configuration;
+
+/**
+ Gets the current configuration in use by the Parse SDK.
+
+ @return The current configuration in use by the SDK. Returns nil if the SDK has not been initialized yet.
+ */
++ (ParseClientConfiguration *)currentConfiguration;
 
 /**
  The current application id that was used to configure Parse framework.
--- a/Parse/Parse.m
+++ b/Parse/Parse.m
@@ -11,6 +11,7 @@
 #import "Parse.h"
 #import "ParseInternal.h"
 #import "ParseManager.h"
+#import "ParseClientConfiguration_Private.h"
 #import "PFEventuallyPin.h"
 #import "PFObject+Subclass.h"
 #import "PFOfflineStore.h"
@@ -34,11 +35,7 @@
 @implementation Parse
 
 static ParseManager *currentParseManager_;
-
-static BOOL shouldEnableLocalDatastore_;
-
-static NSString *applicationGroupIdentifier_;
-static NSString *containingApplicationBundleIdentifier_;
+static ParseClientConfiguration *currentParseConfiguration_;
 
 + (void)initialize {
     if (self == [Parse class]) {
@@ -46,6 +43,8 @@ static NSString *containingApplicationBundleIdentifier_;
         // Without this call - private categories - will require `-ObjC` in linker flags.
         // By explicitly calling empty method - we can avoid that.
         [PFCategoryLoader loadPrivateCategories];
+
+        currentParseConfiguration_ = [ParseClientConfiguration emptyConfiguration];
     }
 }
 
@@ -54,17 +53,30 @@ static NSString *containingApplicationBundleIdentifier_;
 ///--------------------------------------
 
 + (void)setApplicationId:(NSString *)applicationId clientKey:(NSString *)clientKey {
-    PFConsistencyAssert([applicationId length], @"'applicationId' should not be nil.");
-    PFConsistencyAssert([clientKey length], @"'clientKey' should not be nil.");
-    
-    // Setup new manager first, so it's 100% ready whenever someone sends a request for anything.
-    ParseManager *manager = [[ParseManager alloc] initWithApplicationId:applicationId clientKey:clientKey];
-    [manager configureWithApplicationGroupIdentifier:applicationGroupIdentifier_
-                     containingApplicationIdentifier:containingApplicationBundleIdentifier_
-                               enabledLocalDataStore:shouldEnableLocalDatastore_];
-    currentParseManager_ = manager;
+    currentParseConfiguration_.applicationId = applicationId;
+    currentParseConfiguration_.clientKey = clientKey;
 
-    shouldEnableLocalDatastore_ = NO;
+    [self initializeWithConfiguration:currentParseConfiguration_];
+
+    // This is needed to reset LDS's state in between initializations of Parse. We rely on this in the
+    // context of unit tests.
+    currentParseConfiguration_.localDatastoreEnabled = NO;
+}
+
++ (void)initializeWithConfiguration:(ParseClientConfiguration *)configuration {
+    PFConsistencyAssert(configuration.applicationId.length != 0,
+                        @"You must set your configuration's `applicationId` before calling %s!", __PRETTY_FUNCTION__);
+    PFConsistencyAssert(configuration.clientKey.length != 0,
+                        @"You must set your configuration's `clientKey` before calling %s!", __PRETTY_FUNCTION__);
+    PFConsistencyAssert(![PFApplication currentApplication].extensionEnvironment ||
+                        configuration.applicationGroupIdentifier == nil ||
+                        configuration.containingApplicationBundleIdentifier != nil,
+                        @"'containingApplicationBundleIdentifier' must be non-nil in extension environment");
+
+    ParseManager *manager = [[ParseManager alloc] initWithConfiguration:configuration];
+    [manager startManaging];
+
+    currentParseManager_ = manager;
 
     PFObjectSubclassingController *subclassingController = [PFObjectSubclassingController defaultController];
     // Register built-in subclasses of PFObject so they get used.
@@ -88,19 +100,23 @@ static NSString *containingApplicationBundleIdentifier_;
 
     [currentParseManager_ preloadDiskObjectsToMemoryAsync];
 
-    [[self parseModulesCollection] parseDidInitializeWithApplicationId:applicationId clientKey:clientKey];
+    [[self parseModulesCollection] parseDidInitializeWithApplicationId:configuration.applicationId clientKey:configuration.clientKey];
+}
+
++ (ParseClientConfiguration *)currentConfiguration {
+    return currentParseManager_.configuration;
 }
 
 + (NSString *)getApplicationId {
     PFConsistencyAssert(currentParseManager_,
                         @"You have to call setApplicationId:clientKey: on Parse to configure Parse.");
-    return currentParseManager_.applicationId;
+    return currentParseManager_.configuration.applicationId;
 }
 
 + (NSString *)getClientKey {
     PFConsistencyAssert(currentParseManager_,
                         @"You have to call setApplicationId:clientKey: on Parse to configure Parse.");
-    return currentParseManager_.clientKey;
+    return currentParseManager_.configuration.clientKey;
 }
 
 ///--------------------------------------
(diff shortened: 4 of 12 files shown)

### CODE REVIEW

File: Parse/ParseClientConfiguration.h [hunk 1]
```diff
@@ -0,0 +1,27 @@
...
+ * Copyright (c) 2015-present, Parse, LLC.
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree. An additional grant
+ * of patent rights can be found in the PATENTS file in the same directory.
+ */
+
+#import <Parse/PFConstants.h>
+
```
nlutsenko: Nullability annotate, pretty please. :grin:

Parse/ParseClientConfiguration.h
```diff
@@ -0,0 +1,27 @@
+/**
+ * Copyright (c) 2015-present, Parse, LLC.
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree. An additional grant
+ * of patent rights can be found in the PATENTS file in the same directory.
+ */
+
```
nlutsenko: Maybe add import for Foundation? Since we rely on `NSObject`/`NSCopying`/`NSString` here?

Parse/ParseClientConfiguration.h
```diff
@@ -0,0 +1,27 @@
...
+@interface ParseClientConfiguration : NSObject<NSCopying>
+
+@property (nonatomic, copy) NSString *applicationId;
+@property (nonatomic, copy) NSString *clientKey;
+
+@property (nonatomic, assign, getter=isLocalDatastoreEnabled)BOOL localDatastoreEnabled;
+
+@property (nonatomic, copy) NSString *applicationGroupIdentifier;
+@property (nonatomic, copy) NSString *containingApplicationBundleIdentifier;
+
```
nlutsenko: Mark `init` and `new` as unavailable, forcing initialization purely via designated constructor.

Parse/ParseClientConfiguration.h
```diff
@@ -0,0 +1,27 @@
...
+ * LICENSE file in the root directory of this source tree. An additional grant
+ * of patent rights can be found in the PATENTS file in the same directory.
+ */
+
+#import <Parse/PFConstants.h>
+
+/*!
+ This represents the entirety of a parse configuration
+ */
+@interface ParseClientConfiguration : NSObject<NSCopying>
```
nlutsenko: I feel like all these properties should be `readonly`, right?

Parse/ParseClientConfiguration.h
```diff
@@ -0,0 +1,27 @@
...
+
+@property (nonatomic, copy) NSString *applicationId;
+@property (nonatomic, copy) NSString *clientKey;
+
+@property (nonatomic, assign, getter=isLocalDatastoreEnabled)BOOL localDatastoreEnabled;
```
nlutsenko: Nit: Space before `BOOL`

Parse/ParseClientConfiguration.h
nlutsenko: Maybe create a `ParseMutableClientConfiguration` protocol with mutable types and make all of these here immutable?

Tests/Unit/ExtensionDataSharingMobileTests.m
```diff
@@ -47,7 +47,6 @@ - (void)tearDown {
     [fileManager removeItemAtPath:[[Parse _currentManager].fileManager parseLocalSandboxDataDirectoryPath] error:nil];
 
     [Parse _clearCurrentManager];
-    [Parse _resetDataSharingIdentifiers];
```
nlutsenko: Not sure why you don't need it... Add it back?

Parse/Parse.m
```diff
@@ -110,10 +132,7 @@ + (NSString *)getClientKey {
 + (void)enableDataSharingWithApplicationGroupIdentifier:(NSString *)groupIdentifier {
     PFConsistencyAssert(!currentParseManager_,
                         @"'enableDataSharingWithApplicationGroupIdentifier:' must be called before 'setApplicationId:clientKey'");
-    PFParameterAssert([groupIdentifier length], @"'groupIdentifier' should not be nil.");
```
nlutsenko: This validation doesn't exist anymore at this point. Probably should be brought back.

Parse/Parse.m
```diff
@@ -121,11 +140,6 @@ + (void)enableDataSharingWithApplicationGroupIdentifier:(NSString *)groupIdentif
                                  containingApplication:(NSString *)bundleIdentifier {
     PFConsistencyAssert(!currentParseManager_,
                         @"'enableDataSharingWithApplicationGroupIdentifier:containingApplication:' must be called before 'setApplicationId:clientKey'");
-    PFParameterAssert([groupIdentifier length], @"'groupIdentifier' should not be nil.");
-    PFParameterAssert([bundleIdentifier length], @"Containing application bundle identifier should not be nil.");
```
nlutsenko: 124-125 validation is no longer there, we probably still want it.

Parse/ParseClientConfiguration.h
```diff
@@ -0,0 +1,50 @@
...
+
+@end
+
+/*!
+ This represents the entirety of a parse configuration
+ */
+@interface ParseClientConfiguration : NSObject<NSCopying>
+
++ (instancetype)new NS_UNAVAILABLE;
++ (instancetype)alloc NS_UNAVAILABLE;
```
nlutsenko: I don't think we need these, simply due to the fact that you can't `init` it anyway after allocating... 2 less methods polluting the name space?

Parse/ParseClientConfiguration.h
```diff
@@ -0,0 +1,50 @@
...
++ (instancetype)allocWithZone:(NSZone *)zone NS_UNAVAILABLE;
+- (instancetype)init NS_UNAVAILABLE;
+
+@property (nonatomic, readonly, copy) NSString *applicationId;
```
nlutsenko: Maybe place properties first, since most people care about them...

Parse/ParseClientConfiguration.h
```diff
@@ -0,0 +1,50 @@
...
+@property (nonatomic, readonly, copy) NSString *clientKey;
+
+@property (nonatomic, readonly, assign, getter=isLocalDatastoreEnabled) BOOL localDatastoreEnabled;
```
nlutsenko: Order is as follows (for consistency):
`(nullable/null_unspecified/null_resettable), nonatomic/atomic, assign/strong/copy/retain, (readonly/readwrite), (getter=...), (setter=...)`

richardjrossiii:
Hey all, just added some preliminary APIs for configuring URL sessions & retry attempts that we use in the SDK. Any feedback would be great!

oferRounds:
Hi @richardjrossiii 
Thanks for your very detailed and answer and for your very quick respond!
The changes you've added look great! That will suit exactly what I'm looking for.

When do you think this PR could be merged into master?

Parse/Parse.h
```diff
@@ -67,6 +68,22 @@ NS_ASSUME_NONNULL_BEGIN
 + (void)setApplicationId:(NSString *)applicationId clientKey:(NSString *)clientKey;
```
grantland: I never realized that iOS uses `set` instead of `initialize`. I do like moving towards `initialize` since it's not exactly a setter, but is it weird to have it be named differently?

Also, if we keep `initialize`, we should probably update the abstract of it from `Sets the configuration to be used for parse` to `Initialize Parse with a configuration` or something

Parse/Parse.m
```diff
@@ -54,17 +54,35 @@ + (void)initialize {
...
+
+    if (configuration.applicationGroupIdentifier) {
+        if ([PFApplication currentApplication].extensionEnvironment) {
+            PFConsistencyAssert([configuration.containingApplicationBundleIdentifier length], @"'containingApplicationBundleIdentifier' cannot be nil in extension environment");
+        } else {
+            PFConsistencyAssert(![configuration.containingApplicationBundleIdentifier length], @"'containingApplicationBundleIdentifier' must be nil in non-extension environment");
+        }
+
+        PFConsistencyAssert([PFFileManager isApplicationGroupContainerReachableForGroupIdentifier:configuration.applicationGroupIdentifier],
+                            @"ApplicationGroupContainer is unreachable. Please double check your Xcode project settings.");
```
grantland: IMO most of these checks should go into the builders allocation? Thoughts @nlutsenko @hallucinogen @wangmengyan95?

nlutsenko: I actually agree, that most of these would preferrably fit a configuration object custom setters, rather than here.

richardjrossiii: Sounds good to me. Will update.

Parse/Parse.h
richardjrossiii: My best guess for the reason that we use `set` instead of initialize on iOS - `+initialize` is a method declared by `NSObject`, equivalent to a static contractor in java-land. 

It would have caused some ambiguity for developers if we had a method call like `[Parse initialize:@"appKey" clientKey:@"ck"]`, but we can get around that now by marking `+initialize` as unavailable for use (which wasn't possible when Parse was first created).

grantland: Ah that makes sense, SGTM!

Parse/ParseClientConfiguration.h
```diff
@@ -0,0 +1,54 @@
...
+
+@property (nullable, nonatomic, copy, readwrite) NSString *applicationId;
+@property (nullable, nonatomic, copy, readwrite) NSString *clientKey;
+
+@property (nonatomic, assign, readwrite, getter=isLocalDatastoreEnabled) BOOL localDatastoreEnabled;
+
+@property (nullable, nonatomic, copy, readwrite) NSString *applicationGroupIdentifier;
+@property (nullable, nonatomic, copy, readwrite) NSString *containingApplicationBundleIdentifier;
+
+@property (null_resettable, nonatomic, copy, readwrite) NSURLSessionConfiguration *URLSessionConfiguration;
```
grantland: Should we be publicizing this? I'm afraid of the possibility that we change our networking layer at some point in the future and this configuration becomes obsolete.

What kind of configuration does this API provide? Maybe we should abstract these configurations.

grantland:
Somewhat nitpicky, but `ParseClientConfiguration` seems super long. Would `ParseConfiguration` be confusing with `ParseConfig`?

nlutsenko:
> Somewhat nitpicky, but ParseClientConfiguration seems super long. Would ParseConfiguration be confusing with ParseConfig?

Yup, it's definitely confusing. It's ObjC anyway, so long names are expected...
Plus - this way lets us name the class in the same way across multiple platforms (there is `ParseConfig` already on the Java).

Another potential name that could work is `ParseClientConfig`, but then it's still somewhat confusing compared to `ParseConfig`.

Parse/ParseClientConfiguration.h
nlutsenko: Agreed that purely to make #614 work, we probably shouldn't expose the full session configuration just yet (so sorry @oferRounds), unless there is more need for it, as it could break a lot of things.

Parse/ParseClientConfiguration.m
```diff
@@ -0,0 +1,103 @@
...
+@interface ParseClientConfiguration(Private) <ParseMutableClientConfiguration>
+@end
+
+@implementation ParseClientConfiguration
+
+///--------------------------------------
+#pragma mark - Init
+///--------------------------------------
+
+- (instancetype)initWithConfigurationBlock:(void (^)(id<ParseMutableClientConfiguration>))configurationBlock {
```
nlutsenko: Rename to `initWithBlock:`, as you don't have it in the static constructor.

Parse/ParseClientConfiguration.m
```diff
@@ -0,0 +1,103 @@
...
+    return PFIntegerPairHash(self.applicationId.hash, self.clientKey.hash);
+}
+
+- (BOOL)isEqual:(id)object {
+    if (![object isKindOfClass:[ParseClientConfiguration class]]) {
+        return NO;
+    }
+
+    ParseClientConfiguration *other = object;
+    return ([self.applicationId isEqual:other.applicationId] &&
```
nlutsenko: Looks like this one needs `URLSessionRetryAttempts` comparison.

Parse/ParseClientConfiguration.h
```diff
@@ -0,0 +1,54 @@
...
+@property (nullable, nonatomic, copy, readonly) NSString *applicationId;
+@property (nullable, nonatomic, copy, readonly) NSString *clientKey;
+
+@property (nonatomic, assign, readonly, getter=isLocalDatastoreEnabled) BOOL localDatastoreEnabled;
+
+@property (nullable, nonatomic, copy, readonly) NSString *applicationGroupIdentifier;
+@property (nullable, nonatomic, copy, readonly) NSString *containingApplicationBundleIdentifier;
+
+@property (null_resettable, nonatomic, copy, readonly) NSURLSessionConfiguration *URLSessionConfiguration;
+@property (nonatomic, assign, readonly) NSUInteger URLSessionRetryAttempts;
```
nlutsenko: Maybe rename to `networkRetryAttempts`, due to the fact that usage of `URLSession` is an implementation detail.

Parse/ParseClientConfiguration.m
```diff
@@ -0,0 +1,103 @@
...
+///--------------------------------------
+
+- (instancetype)copyWithZone:(NSZone *)zone {
+    return [ParseClientConfiguration configurationWithBlock:^(ParseClientConfiguration *configuration) {
+        configuration.applicationId = self.applicationId;
+        configuration.clientKey = self.clientKey;
+        configuration.localDatastoreEnabled = self.localDatastoreEnabled;
+        configuration.applicationGroupIdentifier = self.applicationGroupIdentifier;
+        configuration.containingApplicationBundleIdentifier = self.containingApplicationBundleIdentifier;
+    }];
```
nlutsenko: Needs to copy over `URLSessionRetryAttempts`(`networkRetryAttempts`).

Tests/Unit/ExtensionDataSharingTests.m
```diff
@@ -72,10 +74,10 @@ - (void)testEnablingDataSharingFromMainApp {
     [Parse _resetDataSharingIdentifiers];
 
     _testHelper.runningInExtensionEnvironment = YES;
-    XCTAssertThrows([Parse enableDataSharingWithApplicationGroupIdentifier:@"yolo"]);
+    [Parse enableDataSharingWithApplicationGroupIdentifier:@"yolo"];
 
     // Just to make sure that initialization runs smoothly
-    [Parse setApplicationId:[[NSUUID UUID] UUIDString] clientKey:[[NSUUID UUID] UUIDString]];
+    XCTAssertThrows([Parse setApplicationId:[[NSUUID UUID] UUIDString] clientKey:[[NSUUID UUID] UUIDString]]);
```
nlutsenko: This probably will change back if we move the validation to `ParseClientConfiguration` itself.

Parse/ParseClientConfiguration.m
```diff
@@ -0,0 +1,103 @@
...
+@property (nullable, nonatomic, copy, readwrite) NSString *containingApplicationBundleIdentifier;
+
+@property (null_resettable, nonatomic, copy, readwrite) NSURLSessionConfiguration *URLSessionConfiguration;
+@property (nonatomic, assign, readwrite) NSUInteger URLSessionRetryAttempts;
+
+@end
+
+// We must implement the protocol here otherwise clang issues warnings about non-matching property declarations.
+// For some reason if the property declarations are on a separate category, it doesn't care.
+@interface ParseClientConfiguration(Private) <ParseMutableClientConfiguration>
```
nlutsenko: Nit: Missing space before `(Private)`

Parse/ParseClientConfiguration.m
```diff
@@ -0,0 +1,103 @@
...
+ * LICENSE file in the root directory of this source tree. An additional grant
+ * of patent rights can be found in the PATENTS file in the same directory.
+ */
+
+#import "ParseClientConfiguration.h"
+
+#import "PFHash.h"
+#import "PFCommandRunningConstants.h"
+
+@interface ParseClientConfiguration()
```
nlutsenko: Nit: Missing space before `()`

Parse/Internal/ParseManager.m
```diff
@@ -108,17 +108,7 @@ - (instancetype)initWithApplicationId:(NSString *)applicationId clientKey:(NSStr
     _controllerAccessQueue = dispatch_queue_create("com.parse.controller.access", DISPATCH_QUEUE_SERIAL);
     _preloadQueue = dispatch_queue_create("com.parse.preload", DISPATCH_QUEUE_SERIAL);
 
-    _applicationId = [applicationId copy];
-    _clientKey = [clientKey copy];
-
-    return self;
-}
-
-- (void)configureWithApplicationGroupIdentifier:(NSString *)applicationGroupIdentifier
```
nlutsenko: I think we should still split the configuration that happens here into a separate method outside of the constructor, simply due to the fact that the methods calls in this method will actually load data/configure/read from disk, which might be a good idea to move outside of the scope of the initialization. As discussed - a potentially good name is `startManaging` or `start`?

grantland:
>> Somewhat nitpicky, but ParseClientConfiguration seems super long. Would ParseConfiguration be confusing with ParseConfig?

> Yup, it's definitely confusing. It's ObjC anyway, so long names are expected...
Plus - this way lets us name the class in the same way across multiple platforms (there is ParseConfig already on the Java).

> Another potential name that could work is ParseClientConfig, but then it's still somewhat confusing compared to ParseConfig.

Any other potential names @hallucinogen @wangmengyan95 @andrewimm @stanleyw? I really don't want a 24 character classname for non-objc...

richardjrossiii:
@grantland For .NET, it's set up as `ParseClient.Configuration`, and I was thinking a similar naming system would work for Java. Especially when you can explicitly import `ParseClient.Configuration`, and not have to fully qualify it everywhere, in both languages.

Only in objective-c where we don't have subtypes or namespaces do we want a classname like this.

grantland:
Is it weird that we have `ParseClient` and `ParseClient.Configuration` on .NET, but `Parse` and `ParseClientConfiguration` on ObjC? Seems like it should logically be `ParseConfiguration` to match `Parse.Configuration` on Java.

richardjrossiii:
@grantland  `ParseConfiguration` is simply too ambiguous with `ParseConfig` I think.

nlutsenko:
Agreed with the above. `ParseConfiguration` is waay to ambiguous with `ParseConfig`.
In the ideal world - we would have `Parse.Client.Configuration` aka `Namespace.Class.Struct` in Swifty world, but we need to stick to `Parse` as the basic class name in ObjC.
Though we can safely rename it at this point, by using say a typealias. (larger discussion, but throwing this here).

grantland:
I guess I'm fine with it then

nlutsenko:
Pass back when you are ready.

Parse/Internal/ParseClientConfiguration_Private.h
```diff
@@ -0,0 +1,33 @@
...
+@property (nullable, nonatomic, copy, readwrite) NSString *clientKey;
+
+@property (nonatomic, assign, readwrite, getter=isLocalDatastoreEnabled) BOOL localDatastoreEnabled;
+
+@property (nullable, nonatomic, copy, readwrite) NSString *applicationGroupIdentifier;
+@property (nullable, nonatomic, copy, readwrite) NSString *containingApplicationBundleIdentifier;
+
+@property (nonatomic, assign, readwrite) NSUInteger networkRetryAttempts;
+
++ (nonnull instancetype)emptyConfiguration;
```
nlutsenko: Use `NS_ASSUME_NONNULL_*` instead of adding `nonnull` to this method.

Parse/Parse.h
```diff
@@ -67,6 +68,22 @@ NS_ASSUME_NONNULL_BEGIN
...
+ @note Re-setting the configuration after having previously sent requests through the SDK results in undefined behavior.
+
+ @param configuration The new configuration to set for the SDK.
+ */
++ (void)initializeWithConfiguration:(ParseClientConfiguration *)configuration;
+
+/**
+ Gets the current configuration in use by the Parse SDK.
+
+ @returns The current configuration in use by the SDK. Returns nil if the SDK has not been initialized yet.
```
nlutsenko: Nit: `@return`

Parse/Parse.h
```diff
@@ -67,6 +68,22 @@ NS_ASSUME_NONNULL_BEGIN
 + (void)setApplicationId:(NSString *)applicationId clientKey:(NSString *)clientKey;
 
 /**
+ Sets the configuration to be used for parse.
```
nlutsenko: Nit: `...for Parse SDK`

Parse/ParseClientConfiguration.h
```diff
@@ -0,0 +1,123 @@
...
+ It is only usable during the execution of the block passed to `ParseClientConfiguration.+configurationWithBlock:`,
+ during which time you should set your properties on it, similar to the following:
+
+ ```
+ configuration.applicationId = @"<#YOUR APPLICATION ID#>"
+ configuration.clientKey = @"<#YOUR CLIENT KEY#>"
+ configuration.localDatastoreEnabled = true
+ ```
+ */
+@protocol ParseMutableClientConfiguration<NSObject>
```
nlutsenko: Nit: Space before `<NSObject>`, as it looks like we are using that style.

Parse/ParseClientConfiguration.h
```diff
@@ -0,0 +1,123 @@
...
+
+/**
+ The Parse.com client key to configure the SDK with.
+ */
+@property (nullable, nonatomic, copy, readwrite) NSString *clientKey;
+
+/**
+ Whether or not to enable pinning in the SDK.
+
+ The default value is `false`.
```
nlutsenko: Nit: `NO` instead of `false`, since we are not quite swift yet (pun intended).

Parse/ParseClientConfiguration.h
```diff
@@ -0,0 +1,123 @@
...
+
+/*!
+ The `ParseClientConfiguration` represents the local configuration of the SDK to connect to the server with.
+
+ These configurations can be stored, copied, and compared, but cannot be safely changed once the SDK is initialized.
+
+ Use this object to construct a configuration for the SDK in your application, and pass it to
+ `Parse.+initializeWithConfiguration:`.
+ */
+@interface ParseClientConfiguration : NSObject<NSCopying>
```
nlutsenko: Nit: Space before `<NSCopying>`, or leave it as is if clang-format.

Parse/ParseClientConfiguration.h
```diff
@@ -0,0 +1,123 @@
...
+ @warning This property should only be set from inside an extension environment.
+ */
+@property (nullable, nonatomic, copy, readonly) NSString *containingApplicationBundleIdentifier;
+
+/**
+ The maximum number of retry attempts to make upon a failed network request.
+ */
+@property (nonatomic, assign, readonly) NSUInteger networkRetryAttempts;
+
++ (instancetype)configurationWithBlock:(void(^)(id<ParseMutableClientConfiguration> configuration))configurationBlock;
```
nlutsenko: Please double check if it's exported to Swift interface with the intended initializer, and not `ParseClientConfiguration.configurationWithBlock { ... in ... }`

Parse/ParseClientConfiguration.m
```diff
@@ -0,0 +1,125 @@
...
++ (instancetype)configurationWithBlock:(void (^)(id<ParseMutableClientConfiguration>))configurationBlock {
+    return [[self alloc] initWithBlock:configurationBlock];
+}
+
+///--------------------------------------
+#pragma mark - Properties
+///--------------------------------------
+
+- (void)setApplicationId:(NSString *)applicationId {
+    PFConsistencyAssert([applicationId length], @"'applicationId' should not be nil.");
```
nlutsenko: Nit: Use property accessor.

Parse/ParseClientConfiguration.m
```diff
@@ -0,0 +1,125 @@
...
+    _applicationId = [applicationId copy];
+}
+
+- (void)setClientKey:(NSString *)clientKey {
+    PFConsistencyAssert([clientKey length], @"'clientKey' should not be nil.");
```
nlutsenko: Nit: Use property accessor.

Parse/ParseClientConfiguration.m
```diff
@@ -0,0 +1,125 @@
...
+    _clientKey = [clientKey copy];
+}
+
+- (void)setApplicationGroupIdentifier:(NSString *)applicationGroupIdentifier {
+    PFConsistencyAssert(
```
nlutsenko: Do we want to let people specify `nil` group identifier?
I think this needs to be the other way around.

Parse/ParseClientConfiguration.m
```diff
@@ -0,0 +1,125 @@
...
+- (NSUInteger)hash {
+    return PFIntegerPairHash(self.applicationId.hash, self.clientKey.hash);
+}
+
+- (BOOL)isEqual:(id)object {
+    if (![object isKindOfClass:[ParseClientConfiguration class]]) {
+        return NO;
+    }
+
+#define EQUAL_OBJECTS(a, b) ((a) == (b) || [(a) isEqual:(b)])
```
nlutsenko: Can we avoid doing this? Inline defines are quite sad.
There is `PFObjectUtilities.+isObject:equalToObject:` for this very reason :grin:

Tests/Unit/ClientConfigurationTests.m
```diff
@@ -0,0 +1,92 @@
...
+@import Foundation;
+
+#import "PFTestCase.h"
+#import "ParseClientConfiguration.h"
+#import "ParseClientConfiguration_Private.h"
+
+@interface ClientConfigurationTests : PFTestCase
+@end
+
+@implementation ClientConfigurationTests
```
nlutsenko: It looks like we are missing coverage on properties that apply when extensions data sharing is in use. Also, nitpicking - but maybe change tests name to `ParseClientConfigurationTests`.

nlutsenko:
LGTM, minus nits. Awesome work so far though!

Parse/Internal/ParseManager.m
```diff
@@ -108,38 +108,32 @@ - (instancetype)initWithApplicationId:(NSString *)applicationId clientKey:(NSStr
...
+- (void)startManaging {
     // Migrate any data if it's required.
     [self _migrateSandboxDataToApplicationGroupContainerIfNeeded];
 
     // TODO: (nlutsenko) Make it not terrible!
     [[self.persistenceController getPersistenceGroupAsync] waitForResult:nil withMainThreadWarning:NO];
 
-    if (localDataStoreEnabled) {
-        PFOfflineStoreOptions options = (_applicationGroupIdentifier ?
+    if (_configuration.localDatastoreEnabled) {
```
nlutsenko: Nit: Use `self.configuration`?

Parse/Parse.m
```diff
@@ -54,17 +53,25 @@ + (void)initialize {
...
-                     containingApplicationIdentifier:containingApplicationBundleIdentifier_
-                               enabledLocalDataStore:shouldEnableLocalDatastore_];
-    currentParseManager_ = manager;
+    currentParseConfiguration_.applicationId = applicationId;
+    currentParseConfiguration_.clientKey = clientKey;
+
+    [self initializeWithConfiguration:currentParseConfiguration_];
+
+    // TODO(richardross): Investigate why this is necessary.
+    currentParseConfiguration_.localDatastoreEnabled = NO;
```
nlutsenko: Try removing this, if it's too much work - leave as is.

Parse/Parse.m
```diff
@@ -123,25 +133,21 @@ + (void)enableDataSharingWithApplicationGroupIdentifier:(NSString *)groupIdentif
...
 
-    applicationGroupIdentifier_ = groupIdentifier;
-    containingApplicationBundleIdentifier_ = bundleIdentifier;
+    currentParseConfiguration_.applicationGroupIdentifier = groupIdentifier;
+    currentParseConfiguration_.containingApplicationBundleIdentifier = bundleIdentifier;
 }
 
 + (NSString *)applicationGroupIdentifierForDataSharing {
-    return applicationGroupIdentifier_;
+    return currentParseConfiguration_.applicationGroupIdentifier;
```
nlutsenko: This should go through `currentManager_` if it's initialized.

Parse/ParseClientConfiguration.h
```diff
@@ -0,0 +1,123 @@
...
+ configuration.clientKey = @"<#YOUR CLIENT KEY#>"
+ configuration.localDatastoreEnabled = true
+ ```
+ */
+@protocol ParseMutableClientConfiguration <NSObject>
+
+/**
+ The Parse.com application id to configure the SDK with.
+ */
+@property (nullable, nonatomic, copy, readwrite) NSString *applicationId;
```
nlutsenko: Nit: Kill `readwrite` on all of these.

Parse/Parse.m
```diff
@@ -54,17 +53,25 @@ + (void)initialize {
...
+    currentParseConfiguration_.clientKey = clientKey;
+
+    [self initializeWithConfiguration:currentParseConfiguration_];
+
+    // TODO(richardross): Investigate why this is necessary.
+    currentParseConfiguration_.localDatastoreEnabled = NO;
+}
 
-    shouldEnableLocalDatastore_ = NO;
++ (void)initializeWithConfiguration:(ParseClientConfiguration *)configuration {
```
nlutsenko: This should have a validation on `applicationId`, `clientKey` not being `nil`, or rather not having 0 length.

Parse/ParseClientConfiguration.h
```diff
@@ -0,0 +1,123 @@
...
+@property (nullable, nonatomic, copy, readonly) NSString *containingApplicationBundleIdentifier;
+
+/**
+ The maximum number of retry attempts to make upon a failed network request.
+ */
+@property (nonatomic, assign, readonly) NSUInteger networkRetryAttempts;
+
++ (instancetype)configurationWithBlock:(void (^)(id<ParseMutableClientConfiguration> configuration))configurationBlock;
+
++ (instancetype) new NS_UNAVAILABLE;
```
nlutsenko: Nit: Remove space before `new`.

Parse/ParseClientConfiguration.h
```diff
@@ -0,0 +1,123 @@
...
+ @warning This property should only be set from inside an extension environment.
+ */
+@property (nullable, nonatomic, copy, readonly) NSString *containingApplicationBundleIdentifier;
+
+/**
+ The maximum number of retry attempts to make upon a failed network request.
+ */
+@property (nonatomic, assign, readonly) NSUInteger networkRetryAttempts;
+
++ (instancetype)configurationWithBlock:(void (^)(id<ParseMutableClientConfiguration> configuration))configurationBlock;
```
nlutsenko: Nit: This method feels lonely without documentation.

Parse/ParseClientConfiguration.m
```diff
@@ -0,0 +1,121 @@
...
+                        [PFFileManager isApplicationGroupContainerReachableForGroupIdentifier:applicationGroupIdentifier],
+                        @"ApplicationGroupContainer is unreachable. Please double check your Xcode project settings.");
+
+    _applicationGroupIdentifier = [applicationGroupIdentifier copy];
+}
+
+- (void)setContainingApplicationBundleIdentifier:(NSString *)containingApplicationBundleIdentifier {
+    PFConsistencyAssert([PFApplication currentApplication].extensionEnvironment,
+                        @"'containingApplicationBundleIdentifier' cannot be set in non-extension environment");
+
```
nlutsenko: Nit: Remove extra space.

Parse/ParseClientConfiguration.m
```diff
@@ -0,0 +1,121 @@
...
+    PFConsistencyAssert(containingApplicationBundleIdentifier.length,
+                        @"'containingApplicationBundleIdentifier' should not be nil.");
+
```
nlutsenko: Nit: Remove extra space.

Tests/Unit/ParseClientConfigurationTests.m
```diff
@@ -0,0 +1,92 @@
...
+
+#import "PFTestCase.h"
+#import "ParseClientConfiguration.h"
+#import "ParseClientConfiguration_Private.h"
+
+@interface ParseClientConfigurationTests : PFTestCase
+@end
+
+@implementation ParseClientConfigurationTests
+
```
nlutsenko: Can we add tests for asserts on setters as well as extensions APIs?

nlutsenko:
Only minor nits and request for moar tests.

Parse/Internal/ParseManager.m
```diff
@@ -108,38 +108,32 @@ - (instancetype)initWithApplicationId:(NSString *)applicationId clientKey:(NSStr
...
     return self;
 }
 
-- (void)configureWithApplicationGroupIdentifier:(NSString *)applicationGroupIdentifier
-                containingApplicationIdentifier:(NSString *)containingApplicationIdentifier
-                          enabledLocalDataStore:(BOOL)localDataStoreEnabled {
-    _applicationGroupIdentifier = [applicationGroupIdentifier copy];
-    _containingApplicationIdentifier = [containingApplicationIdentifier copy];
+///--------------------------------------
+#pragma mark - Offline Store
```
nlutsenko: Not sure if this is the appropriate place for it, as `startManaging` doesn't quite relate to Offline Store only.

Parse/Parse.m
nlutsenko: Can we have this validation, please?

Tests/Unit/ParseClientConfigurationTests.m
nlutsenko: Tests for extension APIs still pending, since there is a logic difference between enabling data sharing via Parse vs ParseClientConfiguration.

Parse/ParseClientConfiguration.m
```diff
@@ -0,0 +1,131 @@
...
+    self = [super init];
+    if (!self) return nil;
+
+    _networkRetryAttempts = PFCommandRunningDefaultMaxAttemptsCount;
+
+    return self;
+}
+
+- (instancetype)initWithBlock:(void (^)(id<ParseMutableClientConfiguration>))configurationBlock {
+    self = [super init];
```
nlutsenko: Maybe invoke `-initEmpty` here and declare that as internal designated initializer (class extension in the .m file). Don't feel strongly about it, tho.

nlutsenko:
One small nit, otherwise - it's perfect!

oferRounds:
Hi @richardjrossiii 
Sorry for the question, I'm just not completely sure - if I update my Parse SDK using CocoaPods, will this branch be included?

nlutsenko:
@oferRounds No, not yet, this is going to be included in `1.12.0` release (per milestone on the Pull Request).

oferRounds:
@richardjrossiii - cool, thanks!

PR #572, opened 2015-11-17
Fixed warning in SQLiteDatabaseResult.

richardjrossiii:
Introduced in commit 9757503, simple type-cast warning. Only seems to appear when compiling for testing though.


nlutsenko:
surewhynot

Issue #32, opened 2015-08-17
LDS Performance Improvements


richardjrossiii:
This is a master issue for all things related to local datastore performance improvements.
- [x] Make LDS stop checkpointing and making copies of objects for mutable containers (#102)
- [ ] Make LDS query matching more efficient by not loading every object into memory for a specified pin.
- [ ] Make LDS use `PFObjectState` and `PFUserState` instead of `PFObject` and `PFUser`, to reduce the amount of locking required for every LDS operation.
- [ ] Improve LDS `PFRelation` tracking.
- [ ] Update LDS database schema to allow for query matching in the database layer itself.

For anyone reaching this issue as a result of currently bad LDS performance here's our current recommendations:
- Keep your object schemas as small as possible - the faster it is to parse them from JSON, the better.
- Relations are currently inefficient. If you can avoid them, do so.
- Minimize usage of any values of the following classes:
  - NSArray
  - NSDictionary
  - PFACL
  - PFGeoPoint
- Whenever possible, use smaller pins. Because LDS loads all objects in a single pin before evaluating a query, the smaller the pin is, the better.

rawrmaan:
+1, LDS performance needs lots of work. I'm glad you're finally publicly acknowledging this. See my angry bug report from April 2015: https://developers.facebook.com/bugs/1575754175998187

FreudGit:
+1:      2.3 sec to get one row on a table of 800 objects...  Ouch.

RickDT:
:+1:

FreudGit:
Hi

Sorry to came back.  Do you think one of these point can be associate to a milestone?

Maybe not clean, but a PFQuery with a whereKey pointing to a String can be filtered direct from the DB (json), without unwrap.

Just with this, Query time will be 99% smaller.

Just my 2 cents

nlutsenko:
Hey @FreudGit, absolutely!
I've just attached a PR to this that implements the first bullet point in that list.
From our testing - it should improve performance significantly.

But you are right - we should attach more to milestones. Making it happen...

FreudGit:
WOW, thanks for the fast reply.

PR?  This was a bit to technical for me.  But i'm happy if we can target and see better performance in the next release\weeks. :+1:

nlutsenko:
PR is a shorthand for Pull Request

[...]

PR #591, opened 2015-11-21
Improve performance of string from date creation.

nlutsenko:
This improves creation of string from dates by ~5 times, by using sqlite directly instead of generic NSDateFormatter.
As a result of this improvement:
- 5-10% win in LDS performance on pinning (I am serious, there is a huge ton of time spent here)
- huge performance gain on encoding any JSON that has our dates in it, say for any network request

Contributes to #32, as it improves LDS perf.

Now, onto notes - I intentionally am not resetting the binding on the statements,
as there is slight performance gain there (about 5% compared to resetting them every time).
Couldn't come up with implications on not doing this, as I feel it's safe to do it in this use-case.

### DIFF
--- a/Parse/Internal/PFDateFormatter.m
+++ b/Parse/Internal/PFDateFormatter.m
@@ -10,21 +10,19 @@
 #import "PFDateFormatter.h"
 
 #import <sqlite3.h>
+#import <sys/time.h>
 
 @interface PFDateFormatter () {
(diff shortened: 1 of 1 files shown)


richardjrossiii:
Resetting is safe to not do because we're only binding primitives (aka `double` here). Those will be directly copied into the statement's internal buffer. Love it!

PR #592, opened 2015-11-21
Fix potential crash in PFOfflineObjectEncoder.

nlutsenko:
Use `weak` instead of `assign`.
Also improved constructor by not going through every property setter and assigning default values directly.

### DIFF
--- a/Parse/Internal/PFEncoder.m
+++ b/Parse/Internal/PFEncoder.m
@@ -188,8 +188,8 @@
 
 @interface PFOfflineObjectEncoder ()
 
-@property (nonatomic, assign) PFOfflineStore *store;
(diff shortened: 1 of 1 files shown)


Parse/Internal/PFEncoder.m
```diff
@@ -188,8 +188,8 @@ - (id)encodeParseObject:(PFObject *)object {
 
 @interface PFOfflineObjectEncoder ()
 
-@property (nonatomic, assign) PFOfflineStore *store;
```
richardjrossiii: There should be a static analysis rule for clang that allows for a warning for setting an object type to be assign. Its 2015 people!

PR #600, opened 2015-11-24
builds dynamic framworks

flovilmart:
I have a 'workaround' here to build dynamic frameworks just based on xconfigs, as I needed dynamic frameworks for a swift framework.

This producing the frameworks in build/ios through a shell script (not integrated with the rake bundle).

How do you think you could support that? In the future I will only require dylibs in the project I'm working on.

### DIFF
--- a/Configurations/Parse-iOS-Dynamic.xcconfig
+++ b/Configurations/Parse-iOS-Dynamic.xcconfig
@@ -6,44 +6,18 @@
 // LICENSE file in the root directory of this source tree. An additional grant
 // of patent rights can be found in the PATENTS file in the same directory.
 //
-
+#include "Shared/Platform/iOS.xcconfig"
 #include "Shared/Product/Framework.xcconfig"
+#include "Configurations/Parse-iOS.xcconfig"
+#include "Configurations/Framework-iOS-Dynamic.xcconfig"
 
-PRODUCT_NAME                    = Parse
-IPHONEOS_DEPLOYMENT_TARGET      = 8.0
+
+// Force SDKROOT to the passed var
+SDKROOT = $(SDKROOT)
 // This silence a warning
-
 SDKROOT[sdk=macos*] = iphoneos
-SDKROOT[sdk=iphoneos*]          = iphoneos
-SDKROOT[sdk=iphonesimulator*]   = iphonesimulator
-
-MACH_O_TYPE                     = mh_dylib
-
-ENABLE_BITCODE               = NO
-ENABLE_BITCODE[sdk=iphoneos*]               = YES
-ENABLE_BITCODE[sdk=iphonesimulator*]        = YES
-DEFINES_MODULE = YES
-
-SUPPORTED_PLATFORMS                    = iphonesimulator iphoneos
-
-VALID_ARCHS[sdk=iphoneos*]             = arm64 armv7 armv7s
-VALID_ARCHS[sdk=iphonesimulator*]      = i386 x86_64
-
-ARCHS[sdk=iphoneos*] = arm64 armv7 armv7s
-ARCHS[sdk=iphonesimulator*]      = i386 x86_64
 
 // Breaks the build if intherited $(BUILD_PRODUCTS_DIR)
 FRAMEWORK_SEARCH_PATHS = $PARSE_DIR/build
 
-// Dynamic linking uses different default copy paths
-LD_RUNPATH_SEARCH_PATHS[sdk=iphoneos*]         = $(inherited) '@executable_path/Frameworks' '@loader_path/Frameworks'
-LD_RUNPATH_SEARCH_PATHS[sdk=iphonesimulator*]  = $(inherited) '@executable_path/Frameworks' '@loader_path/Frameworks'
-
-DYLIB_INSTALL_NAME_BASE         = @rpath
-
 OTHER_LDFLAGS                   = $(inherited) '-dynamiclib' '-lsqlite3'
-CLANG_MODULES_AUTOLINK          = YES
-
-ONLY_ACTIVE_ARCH                = NO
-
-INFOPLIST_FILE = $(PROJECT_DIR)/$(PRODUCT_NAME)/Resources/Parse-iOS.Info.plist
--- a/Configurations/Parse-iOS-Dynamic.xcconfig
+++ b/Configurations/Parse-iOS-Dynamic.xcconfig
@@ -1,0 +1,49 @@
+//
+// Copyright (c) 2015-present, Parse, LLC.
+// All rights reserved.
+//
+// This source code is licensed under the BSD-style license found in the
+// LICENSE file in the root directory of this source tree. An additional grant
+// of patent rights can be found in the PATENTS file in the same directory.
+//
+
+#include "Shared/Product/Framework.xcconfig"
+
+PRODUCT_NAME                    = Parse
+IPHONEOS_DEPLOYMENT_TARGET      = 8.0
+// This silence a warning
+
+SDKROOT[sdk=macos*] = iphoneos
+SDKROOT[sdk=iphoneos*]          = iphoneos
+SDKROOT[sdk=iphonesimulator*]   = iphonesimulator
+
+MACH_O_TYPE                     = mh_dylib
+
+ENABLE_BITCODE               = NO
+ENABLE_BITCODE[sdk=iphoneos*]               = YES
+ENABLE_BITCODE[sdk=iphonesimulator*]        = YES
+DEFINES_MODULE = YES
+
+SUPPORTED_PLATFORMS                    = iphonesimulator iphoneos
+
+VALID_ARCHS[sdk=iphoneos*]             = arm64 armv7 armv7s
+VALID_ARCHS[sdk=iphonesimulator*]      = i386 x86_64
+
+ARCHS[sdk=iphoneos*] = arm64 armv7 armv7s
+ARCHS[sdk=iphonesimulator*]      = i386 x86_64
+
+// Breaks the build if intherited $(BUILD_PRODUCTS_DIR)
+FRAMEWORK_SEARCH_PATHS = $PARSE_DIR/build
+
+// Dynamic linking uses different default copy paths
+LD_RUNPATH_SEARCH_PATHS[sdk=iphoneos*]         = $(inherited) '@executable_path/Frameworks' '@loader_path/Frameworks'
+LD_RUNPATH_SEARCH_PATHS[sdk=iphonesimulator*]  = $(inherited) '@executable_path/Frameworks' '@loader_path/Frameworks'
+
+DYLIB_INSTALL_NAME_BASE         = @rpath
+
+OTHER_LDFLAGS                   = $(inherited) '-dynamiclib' '-lsqlite3'
+CLANG_MODULES_AUTOLINK          = YES
+
+ONLY_ACTIVE_ARCH                = NO
+
+INFOPLIST_FILE = $(PROJECT_DIR)/$(PRODUCT_NAME)/Resources/Parse-iOS.Info.plist
--- a/Parse.xcodeproj/project.pbxproj
+++ b/Parse.xcodeproj/project.pbxproj
@@ -1574,6 +1574,13 @@
 /* End PBXBuildFile section */
 
 /* Begin PBXContainerItemProxy section */
+		4A8E7F261C0627D300E8AF28 /* PBXContainerItemProxy */ = {
+			isa = PBXContainerItemProxy;
+			containerPortal = 09D33641139C54930098E916 /* Project object */;
+			proxyType = 1;
+			remoteGlobalIDString = 4A8E7F1E1C0627AF00E8AF28;
+			remoteInfo = "Bolts-iOS-Dynamic";
+		};
 		811167461B8402DA003CB026 /* PBXContainerItemProxy */ = {
 			isa = PBXContainerItemProxy;
 			containerPortal = 09D33641139C54930098E916 /* Project object */;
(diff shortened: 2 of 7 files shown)


nlutsenko:
@flovilmart, is there a specific reason why you need a dynamic framework for Swift framework?
One of the things that I would recommend that would probably unblock you is using "CocoaPods", as it already does the deployment via static/dynamic frameworks.

Will review and report few things on the code here in the meantime. I agree that this is a good idea in general, also would make us compatible with Carthage.

nlutsenko:
Code review:
- We already have configurations built-in for most things, so you likely not need to recreate the new ones from scratch, but rather import existing ones and address few options that are explicitly set for dynamic
- Copyrights in all the headers should remain the same aka -> under Parse with BSD license
- Bash file that you put for building/deploying looks good, but if we want to ship this - we probably should simply duplicate one of the configurations in `Rakefile` and build using that

Overall, I would suggest taking a look at https://github.com/BoltsFramework/Bolts-iOS/pull/168 which was contributed recently to Bolts.framework. That one has the same feature built, but for Bolts.framework. You can base the implementation here off of that one.

flovilmart:
I'm building ParseZero, that allows dump from a ParseApp to be preloaded in the app for a great first launch experience, even offline.It leverages the localDatastore and pins.This framework is in swift and meant to be distributed via Carthage.Mind that Parse is not initialized in that framework, but in the host app.Initially I was using the static lib framework and all was fine, even with the Unit Tests.When integrating in an app, I started having duplicate symbols problems as Parse (and Bolts) would be linked statically twice.Also because the symbols are different the Parse in the host app and the ParseZero framework would be different and uninitialized in the Framework.
Maybe there is a workaround, but Mach-O static is not supported with Swift frameworks.
Also, the dylib support would break iOS7 compatibility. Besides that, I don't see any reason not to start packaging it that way.
I also have in WIP a whole ParseToolkit repo to hole the whole Parse suite and build for Carthage distribution.

Envoyé à partir d'Outlook

    _____________________________

nlutsenko:
That's actually the main reason why we don't ship only dynamic framework for iOS - we want to be explicitly compatible with iOS 7.
ParseZero and ParseToolkit sound like an amazing idea!

One of 2 things for shipping this here:
- I can guide you through the whole process on how to make the dynamic framework happen in Parse SDK
- Or we can build this in ourselves with your help :)

Let me know which path is easier and makes more sense for you.

flovilmart:
As for the review, thanks for the tips :)

flovilmart:
I guess I'll be able to manage with the effort that has been made on Bolts already.

flovilmart:
Also in the Bolts projects, I really don't like the Target duplication that occurs because of the dynamic Scheme.
That is really problematic when a new class is added, one has to remember to add it to all targets.
If you don't mind, I'd propose an alternative solution

nlutsenko:
Sure, that sounds good.
Keep in mind that we have 4 targets right now (iOS/OSX/tvOS/watchOS), and it is pretty manageable, so not big of a problem. I agree that it's ugly and adds extra cognitive overhead, but it is ok so far.

flovilmart:
Should I pursue that route?

capture d ecran 2015-11-24 a 19 51 38

nlutsenko:
Haha! Yeah, that works! Awesome find!

flovilmart:
I'm not sure about setting the destinations from the rakefile, but if that works, that is way cleaner :)

nlutsenko:
One thing - I wonder what happens with Configurations here, I mean `Debug`/`Release`...
Will it pass them forward to another target?

This is implicit and crazy, but I guess it works.

flovilmart:
I'll check it out as we need at least to set a different -destination for the universal framework 





On Tue, Nov 24, 2015 at 4:54 PM -0800, "Nikita Lutsenko" <notifications@github.com> wrote:












One thing - I wonder what happens with Configurations here, I mean Debug/Release...

Will it pass them forward to another target?



This is implicit and crazy, but I guess it works.

flovilmart:
Some parameters can be forwarded like `-configuration` others should be in the environment (like `-destination`).

flovilmart:
@nlutsenko i've factored everything into the Rakefile, it produces 3 zips with the proper frameworks in each. Parse-iOS-dylib.zip contains both Parse and Bolts built as Dynamic frameworks (verified with File)

```file build/release/Parse-iOS-dylib/Bolts.framework/Bolts 
build/release/Parse-iOS-dylib/Bolts.framework/Bolts: Mach-O universal binary with 5 architectures
build/release/Parse-iOS-dylib/Bolts.framework/Bolts (for architecture i386):	Mach-O dynamically linked shared library i386
build/release/Parse-iOS-dylib/Bolts.framework/Bolts (for architecture x86_64):	Mach-O 64-bit dynamically linked shared library x86_64
build/release/Parse-iOS-dylib/Bolts.framework/Bolts (for architecture armv7):	Mach-O dynamically linked shared library arm
build/release/Parse-iOS-dylib/Bolts.framework/Bolts (for architecture armv7s):	Mach-O dynamically linked shared library arm
build/release/Parse-iOS-dylib/Bolts.framework/Bolts (for architecture arm64):	Mach-O 64-bit dynamically linked shared library```

```file build/release/Parse-iOS-dylib/Parse.framework/Parse 
build/release/Parse-iOS-dylib/Parse.framework/Parse: Mach-O universal binary with 5 architectures
build/release/Parse-iOS-dylib/Parse.framework/Parse (for architecture i386):	Mach-O dynamically linked shared library i386
build/release/Parse-iOS-dylib/Parse.framework/Parse (for architecture x86_64):	Mach-O 64-bit dynamically linked shared library x86_64
build/release/Parse-iOS-dylib/Parse.framework/Parse (for architecture armv7):	Mach-O dynamically linked shared library arm
build/release/Parse-iOS-dylib/Parse.framework/Parse (for architecture armv7s):	Mach-O dynamically linked shared library arm
build/release/Parse-iOS-dylib/Parse.framework/Parse (for architecture arm64):	Mach-O 64-bit dynamically linked shared library```

**Comment by flovilmart** [comment]
Although, I still required to have a custom build script for Bolts as it appears Bolts-iOS:master doesn't play well with that (some builds are broken while doing rake package:frameworks)

**Comment by flovilmart** [comment]
@nlutsenko also that won't solve carthage compatibility fully as carthage requires all the shared schemes to build as dylibs

File: Configurations/Parse-iOS-Dynamic.xcconfig [hunk 1]
```diff
@@ -0,0 +1,43 @@
...
+
+SUPPORTED_PLATFORMS                    = iphonesimulator iphoneos
+
+VALID_ARCHS[sdk=iphoneos*]             = arm64 armv7 armv7s
+VALID_ARCHS[sdk=iphonesimulator*]      = i386 x86_64
+
+ARCHS[sdk=iphoneos*] = arm64 armv7 armv7s
+ARCHS[sdk=iphonesimulator*]      = i386 x86_64
+
+FRAMEWORK_SEARCH_PATHS = $(PARSE_DIR)/build/
```
nlutsenko: Maybe $(BUILT_PRODUCTS_DIR), so it's independent?

Configurations/Parse-iOS-Dynamic.xcconfig
```diff
@@ -0,0 +1,43 @@
...
+LD_RUNPATH_SEARCH_PATHS[sdk=iphonesimulator*]  = $(inherited) '@executable_path/Frameworks' '@loader_path/Frameworks'
+
+DYLIB_INSTALL_NAME_BASE         = @rpath
+
+OTHER_LDFLAGS                   = $(inherited) '-dynamiclib' '-lsqlite3'
+CLANG_MODULES_AUTOLINK          = YES
+
+ONLY_ACTIVE_ARCH                = NO
+
+INFOPLIST_FILE = $(PROJECT_DIR)/$(PRODUCT_NAME)/Resources/Parse-iOS.Info.plist
```
nlutsenko: This file needs a new line at the end.

Configurations/Parse-iOS-Dynamic.xcconfig
```diff
@@ -0,0 +1,43 @@
...
+PRODUCT_NAME                    = Parse
+IPHONEOS_DEPLOYMENT_TARGET      = 8.0
+// This silence a warning
+//MACOS_DEPLOYMENT_TARGET           = 8.0
+SDKROOT[sdk=macos*]             = iphoneos
+SDKROOT[sdk=iphoneos*]          = iphoneos
+SDKROOT[sdk=iphonesimulator*]   = iphonesimulator
+MACH_O_TYPE                     = mh_dylib
+
+ENABLE_BITCODE = NO
```
nlutsenko: I am somewhat sceptical about bitcode here actually.
Can you verify that the final binary actually has bitcode?

Gemfile.lock
```diff
@@ -69,6 +69,3 @@ DEPENDENCIES
   rake
   xcodeproj
   xcpretty
-
-BUNDLED WITH
```
nlutsenko: Not sure what happened here, intended change?

Parse.xcodeproj/project.pbxproj
```diff
@@ -1574,6 +1574,13 @@
 /* End PBXBuildFile section */
 
 /* Begin PBXContainerItemProxy section */
+		4A8E7F261C0627D300E8AF28 /* PBXContainerItemProxy */ = {
+			isa = PBXContainerItemProxy;
+			containerPortal = 09D33641139C54930098E916 /* Project object */;
+			proxyType = 1;
+			remoteGlobalIDString = 4A8E7F1E1C0627AF00E8AF28;
+			remoteInfo = "Bolts-iOS-dynamic";
```
nlutsenko: Nitpicking: `Bolts-iOS-Dynamic` for consistency?

Gemfile.lock
flovilmart: nope it did that by itself...

Parse.xcodeproj/project.pbxproj
```diff
@@ -5894,6 +5947,110 @@
 			};
 			name = Release;
 		};
+		4A26C3BD1C053C2D00E5F505 /* Debug */ = {
+			isa = XCBuildConfiguration;
+			baseConfigurationReference = 4A26C3B61C053AEE00E5F505 /* Parse-iOS-Dynamic.xcconfig */;
```
nlutsenko: Can we remove all these build settings from here? 
xcconfig files are exactly for this purpose, since it's crazy to merge these.

Rakefile
```diff
@@ -138,6 +156,7 @@ end
 
 namespace :package do
   package_ios_name = 'Parse-iOS.zip'
+  package_ios_dynamic_name = 'Parse-iOS-dylib.zip'
```
nlutsenko: Nitpicking: `Parse-iOS-Dynamic.zip`?

Parse.xcodeproj/project.pbxproj
flovilmart: yup

Parse.xcodeproj/project.pbxproj
flovilmart: shoot... they were added back...

Rakefile
flovilmart: yup

Configurations/Parse-iOS-Dynamic.xcconfig
flovilmart: done

Configurations/Parse-iOS-Dynamic.xcconfig
flovilmart: I'm skeptical about bit code too... https://medium.com/@FredericJacobs/why-i-m-not-enabling-bitcode-f35cd8fbfcc5

flovilmart: So Parse.framework (dynamic) doesn't have bit code, however, my config for Bolts makes BITCODE_ENABLE=YES so what do you prefer?

nlutsenko: There are people who explicitly want bitcode, so I feel enabling bitcode is a better option.

flovilmart: I'll remove that flag then

nlutsenko:
Forgot to note - can we reuse the shared configurations in xcconfig file for Parse-iOS-Dynamic?
There is a lot of useful things like custom warnings and common setup things.

Configurations/Parse-iOS-Dynamic.xcconfig
flovilmart: It doesn't play well with it, for whatever reason :)

flovilmart:
The shared configuration particularly the iOS breaks stuff as it forces the SDKROOT to iphoneos. I've just played with it and don't seem to have any clean workaround

flovilmart:
@nlutsenko btw, the toolchain is ready flovilmart/ParseToolchain

nlutsenko:
@flovilmart Quick question on top of this:
If you use the shared configuration (as you mentioned it resets the SDKROOT), but specify `sdk=iphoneos` or `sdk=iphonesimulator` in xcodebuild arguments, will that work?

flovilmart:
The override of the SDKROOT is because the Parse-iOS-Dynamic is a macos target and it needs to be overridden by default to an iphone SDKROOT (iphoneos or iphonesimulator)
passing the sdk argument is caught by the custom build script and forwarded so it works. Actually you can see that the build script forwards the SDKROOT from the one passed from the xcodebuild ```-xcconfig Configurations/Parse-iOS-Dynamic.xcconfig -workspace Parse.xcworkspace -scheme Parse-iOS -configuration $(CONFIGURATION) -sdk $SDKROOT CONFIGURATION_BUILD_DIR=$(CONFIGURATION_BUILD_DIR) $(ACTION)```

nlutsenko:
And I am confused... why is it a macos target?

flovilmart:
Xcode defines it that way. 
The custom build target is a macosx target that will execute a script.
In the case of the Bolts-iOS target, you run a custom script that builds the framework, therefore you don't really care about the build settings of that target, but still, you force SDKROOT = iphoneos.

In the case of our Parse-iOS-Dynamic, we need to override those as well. The Parse-iOS-Dynamic.xcconfig is use to define the build settings for the target as well as the build settings for the Parse-iOS target.

nlutsenko:
From what I understood:
The settings from Parse-iOS-Dynamic target are used to infer the settings for the Bolts dynamic target?

Here is the confusing part:
- Bolts-iOS-Dynamic - since it simply invokes a build script that is custom, why do we need to pass all these settings? We have Bolts-iOS that actually builds the framework for both iphoneos and iphonesimulator
- Parse-iOS-Dynamic - it should be purely used to define build settings for the target that builds a dynamic Parse.framework.

Not sure about the underlying difference, but iOS supports dynamic framework targets since Xcode 6 (I think the product type is the same as with macos, you are right).


Let's start from the beginning:
If you import `iOS.xcconfig` from `xctoolchain`, and remove most of the settings that are defined in Parse-iOS-Dynamic.xcconfig, what is the problem that you are experiencing? Let's try to fix that and remove duplicate and unrequired configurations as much as possible.

flovilmart:
I've spent hours already trying to integrate with iOS.xcconfig and it breaks the build. 

`$ rake build:ios_dynamic` 

```** BUILD SUCCEEDED **

fatal error: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/lipo: /Users/florent/src/Parse-SDK-iOS-OSX/build/Release-ios-iphoneos/Parse.framework/Parse and /Users/florent/src/Parse-SDK-iOS-OSX/build/Release-ios-iphonesimulator/Parse.framework/Parse have the same architectures (armv7) and can't be in the same fat output file
Failed to lipo iOS framework.```

flovilmart:
@nlutsenko  Check in the last commit, I managed to do some things a bit cleaner, `Configurations/Framework-iOS-Dynamic.xcconfig` should probably go to the toolchain submodule. What do you think?

flovilmart:
@nlutsenko I just rebased and fixed a stupid typo in a file name

nlutsenko:
Superseeded by #717

PR #603, opened 2015-11-25
Add and use block-based API for querying SQLite database.

nlutsenko:
Doesn't really add any performance improvements just yet, but this is a waay better API that will guarantee us thread-safety, since the only access to SQLite should be in these blocks.

### DIFF
--- a/Parse/Internal/LocalDataStore/SQLite/PFSQLiteDatabase.m
+++ b/Parse/Internal/LocalDataStore/SQLite/PFSQLiteDatabase.m
@@ -177,7 +177,7 @@ int const PFSQLiteDatabaseDatabaseAlreadyClosed = 4;
 #pragma mark - Query Methods
 ///--------------------------------------
 
-- (BFTask *)_executeQueryAsync:(NSString *)sql withArgumentsInArray:(NSArray *)args cachingEnabled:(BOOL)enableCaching {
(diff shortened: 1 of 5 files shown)


Parse/Internal/LocalDataStore/SQLite/PFSQLiteDatabase.m
```diff
@@ -225,9 +225,14 @@ - (BFTask *)executeCachedQueryAsync:(NSString *)sql withArgumentsInArray:(NSArra
     }];
 }
 
-- (BFTask *)executeQueryAsync:(NSString *)sql withArgumentsInArray:(NSArray *)args {
+- (BFTask *)executeQueryAsync:(NSString *)query withArgumentsInArray:(nullable NSArray *)args block:(PFSQLiteDatabaseQueryBlock)block {
     return [BFTask taskFromExecutor:_databaseExecutor withBlock:^id {
-        return [self _executeQueryAsync:sql withArgumentsInArray:args cachingEnabled:NO];
+        BFTask *task = [self _executeQueryAsync:query withArgumentsInArray:args cachingEnabled:NO];
+        return [task continueImmediatelyWithSuccessBlock:^id(BFTask PF_GENERIC(PFSQLiteDatabaseResult *)*task) {
+            id result = block(task.result);
```
richardjrossiii: If `block` throws during execution, we end up with statements not being properly closed. If we ever turn on query caching, we get into a really bad state here.

PR #605, opened 2015-11-25
Batch pointer fetch by 25 objects on querying from LDS.

nlutsenko:
Contributes to #32 by improving performance about ~20% compared to previous one on basic queries.
(using a benchmark with 1000 objects in LDS)

Batch pointer fetch by 25 objects at once, instead of running 25 separate queries.
We can do more in this space by parallelizing the fetch further, but this is the first step into this direction.

Depends on #603

### DIFF
--- a/Parse/Internal/LocalDataStore/OfflineStore/PFOfflineStore.m
+++ b/Parse/Internal/LocalDataStore/OfflineStore/PFOfflineStore.m
@@ -520,30 +520,34 @@ static int const PFOfflineStoreMaximumSQLVariablesCount = 999;
             @strongify(self);
             PFConstraintMatcherBlock matcherBlock = [self.offlineQueryLogic createMatcherForQueryState:queryState user:user];
 
-            BFTask *checkAllObjectsTask = [BFTask taskWithResult:nil];
-            for (NSString *uuid in task.result) {
-                __block PFObject *object = nil;
-
-                checkAllObjectsTask = [[[[checkAllObjectsTask continueWithSuccessBlock:^id(BFTask *task) {
-                    return [self _getPointerAsyncWithUUID:uuid database:database];
-                }] continueWithSuccessBlock:^id(BFTask *task) {
-                    object = task.result;
-                    return [self fetchObjectLocallyAsync:object database:database];
-                }] continueWithSuccessBlock:^id(BFTask *task) {
-                    if (!object.dataAvailable) {
-                        return @NO;
+            BFTask *checkAllTask = [BFTask taskWithResult:nil];
+            NSArray *uuidBatches = [PFInternalUtils arrayBySplittingArray:task.result withMaximumComponentsPerSegment:64];
+            for (NSArray *uuids in uuidBatches) {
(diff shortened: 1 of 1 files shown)


nlutsenko:
cc @freudgit checkout out this branch and test it, let me know how it goes.

FreudGit:
Sure, tomorrow first thing!
How to checkout a branch using pod?

nlutsenko:
Something like:
```ruby
pod 'Parse', :git => 'git@github.com:ParsePlatform/Parse-SDK-iOS-OSX.git', :branch => 'nlutsenko.lds.query.batch'
```

FreudGit:
@nlutsenko From my iPhone 6s.  Look like we have a 15% performance improve. (i'll have a iPhone 5 soon to test\get more interesting result)

one thing strange(not only releted to this specific branch)= Look at the code here:  When i call getFullContent_toLocalDataStore, ALL next call to get result are near 0.1 sec.  If I open the app and ask for result direct, result are near .45. Look like pinAllInBackground keep all in memory :)

https://docs.google.com/spreadsheets/d/1QSRgTNXMdS4gNC62yA_zfNp4IhdY1SyUYD3xwb_vM1Q/edit?usp=sharing

FreudGit:
@nlutsenko iPhone 5 = 2.2s(this branch) versus 2.5s.(1.10.0)

nlutsenko:
W00t 15% perf improvement. 
The part about in memory - all fetches are going to be almost instant, since we have single object instance and as long as you have a point to the object - we cache the fact that we fully loaded the object from LDS.

Parse/Internal/LocalDataStore/OfflineStore/PFOfflineStore.m
```diff
@@ -499,52 +495,59 @@ - (BFTask *)findAsyncForQueryState:(PFQueryState *)queryState
...
+                NSString *uuid = [result stringForColumnIndex:0];
+                [uuids addObject:uuid];
+            }
+            return uuids;
+        }] continueWithSuccessBlock:^id(BFTask PF_GENERIC(NSArray *)*task) {
+            @strongify(self);
+            PFConstraintMatcherBlock matcherBlock = [self.offlineQueryLogic createMatcherForQueryState:queryState user:user];
+
+            BFTask *checkAllTask = [BFTask taskWithResult:nil];
+            NSArray *uuidBatches = [PFInternalUtils arrayBySplittingArray:task.result withMaximumComponentsPerSegment:25];
```
richardjrossiii: Not 100% sure why we're splitting this here instead of doing it in one big batch. Care to clue me in? 

We still only have one query that can go through in the DB at a time, why have extra context switches for smaller batch sizes?

Just seems like needless work, especially when `continueWIthSuccessBlock:` will always run on the same thread anyway...

The only rationalization I can think of is maybe if we wanted to add progress reporting it could be useful, eventually.

nlutsenko: We split to make sure we do not allocate a lot of memory. Will change the limit to 64 to keep it at power of 2.

PR #615, opened 2015-12-02
Update all of the public documentation to new style.

nlutsenko:
This change is required to power our new api ref generation.
- Replaced `/*!` with `/**`
- Removed `@abstract`, `@discussion` and converted `@returns` into `@return`
- Updated all cross references from `<[Class method]>` to `Class.+method` style
- Cleaned up and fixed few missing/wrong pieces of documentation.

The new style is as follows:
```objc
/**
 Short description of what the method does.

 Very long description that precisely describes on what we are doing with a cross reference link to `AnotherClass.+someRandomMethod`.

 @warning Some warning. // (optional)
 @note Some extra note. // (optional)

 @param argument This is a description of what is expected of this argument.

 @return Description of return value.
*/
- (BOOL)methodNameWithArgument:(id)argument;
```

### DIFF
--- a/Parse/Internal/ACL/DefaultACLController/PFDefaultACLController.h
+++ b/Parse/Internal/ACL/DefaultACLController/PFDefaultACLController.h
@@ -30,14 +30,14 @@ NS_ASSUME_NONNULL_BEGIN
 /// @name Default ACL
 ///--------------------------------------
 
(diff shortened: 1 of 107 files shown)


richardjrossiii:
veryseriouscodereview.gif

PR #616, opened 2015-12-02
Mark +parseClassName as required method.

nlutsenko:
All other methods of PFSubclassing are optional and the default implementation is provided by PFObject.

### DIFF
--- a/Parse/PFSubclassing.h
+++ b/Parse/PFSubclassing.h
@@ -23,6 +23,15 @@ NS_ASSUME_NONNULL_BEGIN
  */
 @protocol PFSubclassing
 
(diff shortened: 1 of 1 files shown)


Parse/PFSubclassing.h
```diff
@@ -23,6 +23,15 @@ NS_ASSUME_NONNULL_BEGIN
  */
 @protocol PFSubclassing
 
+@required
```
richardjrossiii: Technically a protocol is `@required` by default, so it's not 100% necessary. Its a style thing though, I'll let you decided whether or not to keep it.

PR #629, opened 2015-12-07
Fix memory leak in PFSQLiteDatabase.

### LINKED ISSUE #626: Memory Leak in iOS Parse SDK v1.10.0
Gentlefolk,

I was using Instruments to check a prototype app for memory leaks. It is linked against the released Parse iOS SDK v1.10.0. Here is a screenshot of Instruments fingering a retain cycle deep in the core of the Parse SDK.

screen shot 2015-12-05 at 08 58 31

As this is a simple app, I'm happy to instrument it however you guys wish to help find what is likely a problem.

Anon,
Andrew

nlutsenko:
There is a leak here, we are referencing an ivar, which is implicitly referencing `self` inside a `BFExecutor`, which is also retained by `self`.
Fixed simply by using a shared queue pointer, which eliminates reference to `self`. Verified via Instruments.app both on iPhone 6S and Simulator.

Fixes #626.


richardjrossiii:
Yep. Good catch. We really should turn on the warning for implicit capture of self...

PR #631, opened 2015-12-08
Use "CFBundleShortVersionString" to get app version

jrmsklar:
The `PFInstallation` class is currently using the key `kCFBundleVersionKey` to get the app version number. Using this key retrieves the value for the "Bundle Version" key in the application's `Info.plist` file, which represents the build number as specified in General tab of the main target:

screen shot 2015-12-08 at 11 47 21 am

Using the key "CFBundleShortVersionString" will retrieve the value for the "Bundle versions string, short" in the `Info.plist` file, which represents the version number as specified in the General tab of the main target:

screen shot 2015-12-08 at 11 51 46 am

This PR updates the `PFInstallation` class to use the key "CFBundleShortVersionString" instead of `kCFBundleVersionKey` when getting the app version.

### DIFF
--- a/Parse/PFInstallation.m
+++ b/Parse/PFInstallation.m
@@ -278,7 +278,7 @@ static NSSet *protectedKeys;
 - (void)_updateVersionInfoFromDevice {
     NSDictionary *appInfo = [NSBundle mainBundle].infoDictionary;
     NSString *appName = appInfo[(__bridge NSString *)kCFBundleNameKey];
-    NSString *appVersion = appInfo[(__bridge NSString *)kCFBundleVersionKey];
+    NSString *appVersion = appInfo[@"CFBundleShortVersionString"];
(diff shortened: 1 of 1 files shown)


nlutsenko:
Hey @jrmsklar, thanks for the PR.

Unfortunately, usage of `CFBundleVersion` is very much intended to support the flow of beta/internal testing, where the final production version would have identical `CFBundleShortVersionString`.
This proved to be quite useful and when I am using Parse personally - I would very much prefer having a build version, so I can track issues directly against the code that was shipped to beta testers.

The mere difference between the two is the fact that `CFBundleShortVersionString` is intended to be surfaced to users/pr, where the `CFBundleVersion`, as you said - is a build number, which is helpful to engineers/developers.

P.S. You can add your own `appVersionString` to the Installation table to workaround this.

jrmsklar:
I see why the usage of `CFBundleVersion` is intended. Thanks for the explanation @nlutsenko!

PR #647, opened 2015-12-11
Add unavailable attribute and a warning for using NetworkActivityIndicatorManager on OS X.

nlutsenko:


richardjrossiii:
surewhynot

PR #659, opened 2015-12-16
Documentation update

richardgroves:
Removed a line about a callback on the method version that doesn't have a callback.

Should something be added about the circumstances under which the BFTask completes/errors?


nlutsenko:
Good catch! Thanks!
There is no need to mention a callback here, as continuation of the task is going to be called regardless.

PR #665, opened 2015-12-17
Inject parse api server url instead of referencing a global constant.

nlutsenko:
This also adds moves the `/1` logic to the constant itself, instead of hard-coding it on URLRequest creation time. 
Depends on #664


richardjrossiii:
So in my fork (https://github.com/richardjrossiii/Parse-SDK-iOS-OSX/tree/richardross.initialization.serverurl), I decoupled all of the request constructor parameters to be dataSource provided via configuration, I was wondering your thoughts on that solution? I think it provides a much cleaner API in the long-term, especially if we want to have more networking options to be able to be set by the end user (e.g. custom `NSURLSession` support, custom `NSURLProtocol` support, etc.).

nlutsenko:
Talked offline, but the idea is to fully inject `ParseClientConfiguration` all the way down to the CommandRunner, due to the fact that we don't want to allow changing of it at any time as well as due to it being an immutable model.

PR #666, opened 2015-12-17
Cleanup tests and convert many of them to pure unit tests.

nlutsenko:
This PR cleanes up and converts many tests to be pure unit, meaning they don't need a Parse to be initialized, but rather depend on the injected configuration/dependencies to run.


nlutsenko:
Whee, ~~order~~ issue №666! Just in time for SW: VII.

PR #668, opened 2015-12-17
Make ParseModuleCollection initialization not dispatch_async multiple times.

richardjrossiii:
By lifting the `dispatch_async` outside of the module collection enumeration, it allows for fewer local variable caputres, improving perfromance, as well as not requiring multiple flushes of the main dispatch queue to execute.


nlutsenko:
surewhynot

PR #675, opened 2015-12-21
Inject user controller into UserAuthenticationController.

nlutsenko:
Don't get global via `PFUser.+userController`, but inject it via `dataSource`.

### DIFF
--- a/Parse/Internal/BFTask+Private.h
+++ b/Parse/Internal/BFTask+Private.h
@@ -22,19 +22,19 @@
 
 @interface BFTask (Private)
 
-- (instancetype)continueAsyncWithBlock:(BFContinuationBlock)block;
-- (instancetype)continueAsyncWithSuccessBlock:(BFContinuationBlock)block;
+- (BFTask *)continueAsyncWithBlock:(BFContinuationBlock)block;
+- (BFTask *)continueAsyncWithSuccessBlock:(BFContinuationBlock)block;
 
 - (BFTask *)continueImmediatelyWithBlock:(BFContinuationBlock)block;
 - (BFTask *)continueImmediatelyWithSuccessBlock:(BFContinuationBlock)block;
 
-- (instancetype)continueWithResult:(id)result;
-- (instancetype)continueWithSuccessResult:(id)result;
+- (BFTask *)continueWithResult:(id)result;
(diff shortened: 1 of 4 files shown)


Parse/Internal/User/AuthenticationProviders/Controller/PFUserAuthenticationController.m
```diff
@@ -113,12 +113,12 @@ - (void)unregisterAuthenticationDelegateForAuthType:(NSString *)authType {
...
 
-- (BFTask *)logInUserAsyncWithAuthType:(NSString *)authType authData:(NSDictionary *)authData {
+- (BFTask PF_GENERIC(PFUser *)*)logInUserAsyncWithAuthType:(NSString *)authType
+                                                  authData:(NSDictionary PF_GENERIC(NSString *, NSString *)*)authData {
     //TODO: (nlutsenko) Make it fully async.
-    //TODO: (nlutsenko) Inject `PFUserController` here.
     PFUser *currentUser = [PFUser currentUser];
     if (currentUser && [PFAnonymousUtils isLinkedWithUser:currentUser]) {
-        if ([currentUser isLazy]) {
+        if (currentUser.isLazy) {
```
richardjrossiii: Shouldn't it be `.lazy`?

Parse/Internal/BFTask+Private.h
```diff
@@ -22,19 +22,19 @@
 
 @interface BFTask (Private)
 
-- (instancetype)continueAsyncWithBlock:(BFContinuationBlock)block;
-- (instancetype)continueAsyncWithSuccessBlock:(BFContinuationBlock)block;
+- (BFTask *)continueAsyncWithBlock:(BFContinuationBlock)block;
```
richardjrossiii: Separate PR? I don't feel strongly however.

Parse/Internal/User/AuthenticationProviders/Controller/PFUserAuthenticationController.m
nlutsenko: I thought so as well, but property is 'isLazy'

Parse/Internal/BFTask+Private.h
nlutsenko: Nah, too much work.

PR #678, opened 2015-12-22
Deprecate target:selector: asynchronous methods in all public classes.

nlutsenko:
This moves all the declarations into class extensions (categories) as well as adds `deprecated` attribute to all methods that end with `...target:selector:`. These methods are pretty much useless and are deprecated in favor of either block-based callbacks or task-based approach.

### DIFF
--- a/Parse/PFCloud+Deprecated.h
+++ b/Parse/PFCloud+Deprecated.h
@@ -1,0 +1,37 @@
+/**
+ * Copyright (c) 2015-present, Parse, LLC.
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree. An additional grant
+ * of patent rights can be found in the PATENTS file in the same directory.
+ */
+
+#import <Parse/PFCloud.h>
+#import <Parse/PFConstants.h>
+
+NS_ASSUME_NONNULL_BEGIN
+
+@interface PFCloud (Deprecated)
+
+/**
+ Calls the given cloud function *asynchronously* with the parameters provided
+ and then executes the given selector when it is done.
+
+ @param function The function name to call.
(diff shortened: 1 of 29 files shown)


Parse/PFCloud+Deprecated.h
```diff
@@ -0,0 +1,37 @@
...
+ @param selector The selector to call when the function call finished.
+ It should have the following signature: `(void)callbackWithResult:(id)result error:(NSError *)error`.
+ Result will be `nil` if error is set and vice versa.
+ 
+ @deprecated Please use `PFCloud.+callFunctionInBackground:withParamaters:` instead.
+ */
++ (void)callFunctionInBackground:(NSString *)function
+                  withParameters:(nullable NSDictionary *)parameters
+                          target:(nullable id)target
+                        selector:(nullable SEL)selector PARSE_DEPRECATED("Please use `PFCloud.+callFunctionInBackground:withParamaters:` instead.");
```
richardjrossiii: Typo. Should be `PFCloud.+callFunctionInBackground:withParameters:`.

Parse/PFFile.h
```diff
@@ -26,7 +26,7 @@ NS_ASSUME_NONNULL_BEGIN
 ///--------------------------------------
 
 - (instancetype)init NS_UNAVAILABLE;
-+ (instancetype)new NS_UNAVAILABLE;
++ (instancetype) new NS_UNAVAILABLE;
```
richardjrossiii: Intention space change? We just killed this for client configuration...

richardjrossiii:
KILL IT WITH FIRE.

Parse/PFFile.h
nlutsenko: Indeed a clang-format bug, fixing now.

PR #686, opened 2015-12-23
Use shields.io badge for Travis-CI

nlutsenko:
Looks like Travis default badge is using any latest branch available, which doesn't quite suite us, since we care only about `master`.
Use the `shields.io` version of it that supports branches.


richardjrossiii:
And somewhat ironically, travis fails on the PR! haha.

PR #689, opened 2015-12-24
Use property accessors via dot-syntax where possible instead of method invocations.

nlutsenko:
Convert the entire codebase to use dot-syntax for properties instead of method invocations.

### DIFF
--- a/Parse/Internal/Analytics/Utilities/PFAnalyticsUtilities.m
+++ b/Parse/Internal/Analytics/Utilities/PFAnalyticsUtilities.m
@@ -17,14 +17,15 @@
     if (!payload || payload == [NSNull null]) {
         payload = @"";
     } else if ([payload isKindOfClass:[NSDictionary class]]) {
-        NSDictionary *dict = payload;
-        NSArray *keys = [[dict allKeys] sortedArrayUsingSelector:@selector(compare:)];
-        NSMutableArray *components = [NSMutableArray arrayWithCapacity:[dict count] * 2];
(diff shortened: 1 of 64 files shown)


richardjrossiii:
My largest issue with this is using dot-syntax for invocations that may have large overhead/create copies of objects `allObjects`/`allKeys`, etc. as those have large performance considerations relative to properties which are truly just a getter, such as `count`, etc.

Looking forward to the swift approach, where a dictionary's `keys` is actually a lazily generated collection I would have no issues, but here in ObjC land I still think some readability is provided by using method invocation syntax.

I'm not going to consider this a blocking issue, and I'll accept this PR for now, but I'd love to have a further conversation on how we can make performance implications more obvious for situations like this (as we have had some issues, LDS in particular) where we hide performance considerations behind "simple" APIs and we end up with large long-term issues.

nlutsenko:
I agree with that, though performance implication of `allObjects` or `allKeys` is quite cheap, compared to other crazy things we do. Most of these are coming from the recent `Foundation` changes, where methods were converted to `readonly` properties. This minimizes the benefit that we get, I agree, but I think using a custom linter or maybe discouraging usage of expensive methods via property by using, say, a custom linter, would be much more beneficial in my mind compared to simply using method invocations in favor of properties.

TL;DR: If there is a property instead of a method - use dot-syntax. To workaround a hidden complexity issues - add a `unavailable` attribute to those properties in a custom header and replace them with methods that explicitly list complexity (aka `getAllKeys` vs `allKeys`).

Issue #604, opened 2015-11-25
Follow Up Issue #563


valeIT:
Seems like I can't reopen Issues... [Original #563 ]

Ok let's start from the beginning. My object models: [Forgot about the reference to PFUser.]

```
class MainModel: PFObject, PFSubclassing {
  @NSManaged var title: String
  @NSManaged var value: Int
  @NSManaged var createdAtLocally: NSDate
  @NSManaged var lastModifiedAtLocally: NSDate
  @NSManaged var goal: Goal
  @NSManaged var colors: ColorModel
  @NSManaged var user: PFUser //PFUser.currentUser()
}

class Goal: PFObject, PFSubclassing {
  @NSManaged var enabled: Bool
  @NSManaged var goal: Int
  @NSManaged var completed: Bool
}

class ColorModel: PFObject, PFSubclassing {
  @NSManaged var text: String
  @NSManaged var background: String
}

//ACL in AppDelegate
    PFACL.setDefaultACL(PFACL(), withAccessForCurrentUser:true)

```

I have a list of `MainModels` shown in a `PFTableViewController` subclass. 

Calling:

```
let query:PFQuery = PFQuery(className:self.parseClassName!)
      query.fromLocalDatastore()
      do{
        if let obj = try query.findObjects() as? [MainModel] {
//          try PFObject.unpinAllObjects()
          for each in obj {
            try each.unpin()
            each.deleteEventually()
            print(each)
          }
        }
      }catch let error as NSError {print(error.debugDescription); print(error.localizedDescription)} catch _ {print("error unpinning")}
    }
```

Or uncommenting`unpinAllObjects` and commenting the loop doesn't work, raises:

```
The operation couldn’t be completed.
Error Domain=Foundation._GenericObjCError Code=0 "(null)"
```
`unpinAllObjectsInBackgroundWithBlock` works when there is an internet connection though...

How they are pinned:
The only objects that get pinned are `MainModel` at creation or when downloaded from the server, nothing else.

nlutsenko:
Any chance you can pinpoint on which line is the exception raised?
Going over code right now and having trouble finding the root of the issue here.

valeIT:
Weird now it doesn't raise anything anymore,no matter what.

[This is done without internet connection.]

It gets the correct MainModel object to be deleted and than calls `- (BFTask *)unpinInBackgroundWithName:(NSString *)name`

And after:
```
- (id)waitForResult:(NSError **)error {
[...]

Issue #563, opened 2015-11-14
Unpinning from LocalDatastore if offline


valeIT:
For what I understand by Design you can't unpin a child object if it has a parent, but you can unpin the parent and it will automatically unpin all the childs, let me know if that's correct.

My object graph is as follow:

MainModel has 2 pointer variables to Colors and Goals (all 2 are custom PFObjects). Colours and Goals don't have any pointer to any object.

when calling myMainModel.unpinInBackground() it doesn't unpin. The success returns true and no error, but when i refresh using loadObjects from PFQueryTableViewController it returns the deleted object as well. Anything I missed? 

If I'm connected to the internet everything works.

valeIT:
Ok let's start from the beginning. My object models: [Forgot about the reference to PFUser.]

```
class MainModel: PFObject, PFSubclassing {
  @NSManaged var title: String
  @NSManaged var value: Int
  @NSManaged var createdAtLocally: NSDate
  @NSManaged var lastModifiedAtLocally: NSDate
  @NSManaged var goal: Goal
  @NSManaged var colors: ColorModel
  @NSManaged var user: PFUser //PFUser.currentUser()
}

class Goal: PFObject, PFSubclassing {
  @NSManaged var enabled: Bool
  @NSManaged var goal: Int
  @NSManaged var completed: Bool
}

class ColorModel: PFObject, PFSubclassing {
  @NSManaged var text: String
  @NSManaged var background: String
}

//ACL in AppDelegate
    PFACL.setDefaultACL(PFACL(), withAccessForCurrentUser:true)

```

I have a list of `MainModels` shown in a `PFTableViewController` subclass. 

Calling:

```
let query:PFQuery = PFQuery(className:self.parseClassName!)
      query.fromLocalDatastore()
      do{
        if let obj = try query.findObjects() as? [MainModel] {
//          try PFObject.unpinAllObjects()
          for each in obj {
            try each.unpin()
            each.deleteEventually()
            print(each)
          }
        }
      }catch let error as NSError {print(error.debugDescription); print(error.localizedDescription)} catch _ {print("error unpinning")}
    }
```

Or uncommenting`unpinAllObjects` and commenting the loop doesn't work, raises:

```
The operation couldn’t be completed.
Error Domain=Foundation._GenericObjCError Code=0 "(null)"
```
unpinAllObjectsInBackgroundWithBlock works when there is an internet connection though...

How they are pinned:
[...]

PR #697, opened 2015-12-29
Fixed potential issue with wrong result type set on pin/unpin methods of PFObject.

nlutsenko:
Due to the fact that some of these methods might return earlier - there could be `nil` result set on `pin`/`unpin` result tasks. This fixes the issue by enforcing generic types across both `PFPinningObjectStore` and `PFObject` implementation.

Contributes to #604, #563


richardjrossiii:
I can't wait till we can have a compiler warning for a lack of generic type, personally.

nlutsenko:
Yeah, that would be extremely helpful.

Issue #674, opened 2015-12-21
Drop support for Xcode 6


richardjrossiii:
Currently, our SDK supports Xcode 6 as well as Swift 1.2, which have both been deprecated since the release of Xcode 7 on September 16, 2015.

At this point, given the clear migration paths that exist from Swift 1.2 to 2.0, we see no reason to maintain support for Xcode 6.x, and plan on dropping support entirely starting January 1, 2016.

This will NOT effect any current SDK releases that exist, as these will purely be SDK changes, not back-end changes.

The most notable client side changes will be removal of the `PF_GENERIC` macro for backwards compatible objective-c generics.

If you have any questions or concerns for this change, please post them below.

PR #698, opened 2015-12-29
Add generic types to PFOfflineStore.

nlutsenko:
This adds generic types to object collections and tasks in PFOfflineStore.


richardjrossiii:
Part of me wants to say that because we haven't had any feedback on #674, we don't need to worry about PF_GENERIC at this point...

nlutsenko:
I like that part of you! Sticking to the plan of eliminating that in `1.13.0`!

PR #699, opened 2015-12-29
Add nullability annotations and generic types to AnalyticsController.

nlutsenko:


Parse/Internal/Analytics/Controller/PFAnalyticsController.m
```diff
@@ -86,7 +86,7 @@ - (BFTask *)trackEventAsyncWithName:(NSString *)name
                                                                              dimensions:encodedDimensions
                                                                            sessionToken:sessionToken];
         return [self.eventuallyQueue enqueueCommandInBackground:command];
-    }] continueWithSuccessResult:@YES];
+    }] continueWithSuccessResult:nil];
```
richardjrossiii: Does this change public API at all?

nlutsenko: Nope. We have the same piece in PFAnalytics itself. The framework I am using: use PFVoid internally, and NSNumber publicly.
So every controller will return a PFVoid as result type of the task, and continueWithSuccessResult:@YES is going to be only in PFAnalytics or PFObject implementation.

PR #700, opened 2015-12-29
Removes -[PFFile getData:] warning when data is cached/available

natanrolnik:
Imagine the following scenario:

I have a `PFFile` as one of his `PFConfig` parameters. Whenever the config is fetched, runs the following code in the completion block:

```obj-c
PFFile *myFile = config[@"myFile"];
if (!myFile.isDataAvailable) {
      [myFile getDataInBackground];
}
```

As most of the time the file will not be changed, it's set to only get the file data (in background, of course) when it's not available in the cache. Somewhere else in the code, where I need the file, I call:

```obj-c
- (NSData *)myData {
      PFFile *myFile = config[@"myFile"];
      if (!myFile.isDataAvailable) {
            return someFallbackData;
      }
      return [PFFile getData];
}
```

In the above case, `getData` will only be called when the data is available locally. It means that PFFile will get from the cache, and there will be no need of downloading the file. Even this way, there is a warning being logged (`Warning: A long-running operation is being executed on the main thread`), because the standard implementation of `-[BFTask waitForResult:]` called automatically `-[BFTask waitForResult:withMainThreadWarning:YES]`.

To conclude: I added a check on get `-[PFFile getData:]` and `-[PFFile getDataStream:]` so it only enables warning if the file is not cached.

WDYT, @nlutsenko @richardjrossiii ?

### DIFF
--- a/Parse/PFFile.m
+++ b/Parse/PFFile.m
@@ -536,7 +536,8 @@ static const unsigned long long PFFileMaxFileSize = 10 * 1024 * 1024; // 10 MB
 }
 
 - (NSData *)getData:(NSError **)error {
-    return [[self getDataInBackground] waitForResult:error];
+    BOOL enableWarning = !self.isDataAvailable;
+    return [[self getDataInBackground] waitForResult:error withMainThreadWarning:enableWarning];
 }
 
 - (NSInputStream *)getDataStream {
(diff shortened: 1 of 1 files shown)


natanrolnik:
Just noticed it's the issue/pull number 700
:tada: Viva open source :tada:

nlutsenko:
Sorry to disappoint you, but this method is still not safe to call from the main thread, as it's loading the data from disk, which is the slowest piece of hardware yet (specifically on mobile devices).

If the file is big (say 10MB) - it would lead to a main thread stall for quite some time.
And if the file is small - opening a file descriptor and reading it is still not optimal operation to perform on the main thread.

nlutsenko:
It sounds like more work, but if you structure your code to be agnostic of whether the data is available for a file and use a task-based getter - it will actually perform much better in the long-term and let you squeeze the most performance for your UI/Event-handling related things, since those could happen only on the main thread.

natanrolnik:
I see. I considered that point of reading data from the disk, but I didn't think it would be considered `A long-running operation` as much as network request.

PR #707, opened 2015-12-30
Remove Crash Reporting references

hramos:
The Starter Projects contained comments that prompted the developer to enable Crash Reporting in their projects. The Parse Crash Reporting tool is has been deprecated in favor of native solutions and will no longer be supported as of March 1, 2016.


nlutsenko:
Awesome! Thank you for cleaning this up!

PR #721, opened 2016-01-05
Replace all (at)name with pragma marks in headers.

nlutsenko:
We were using `@name` only for documentation generation purposes. 
Since our new doc generation supports pragma marks - use those instead for every header.
Starting from this point - we can finally use pragmas across both headers and implementations.


richardjrossiii:
surewhynot

PR #726, opened 2016-01-05
Convert all private ivars of PFUser to properties.

nlutsenko:
Also added generic types to few of them.
Please note - `authData` needs `id` for object type, since we are using `NSNull` for it in few places.

### DIFF
--- a/Parse/Internal/User/PFUserPrivate.h
+++ b/Parse/Internal/User/PFUserPrivate.h
@@ -51,20 +51,17 @@ extern NSString *const PFUserCurrentUserKeychainItemName;
 @end
 
 // Private Properties
-@interface PFUser () {
-    BOOL isCurrentUser;
-    NSMutableDictionary *authData;
-    NSMutableSet *linkedServiceNames;
-    BOOL isLazy;
-}
+@interface PFUser ()
 
-// This earmarks the user as being an "identity" user. This will make saves write through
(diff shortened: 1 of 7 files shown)


Parse/Internal/User/PFUserPrivate.h
```diff
@@ -51,20 +51,17 @@ extern NSString *const PFUserCurrentUserKeychainItemName;
...
+@interface PFUser ()
+
+@property (nonatomic, strong, readonly) NSMutableDictionary<NSString *, id> *authData;
+@property (nonatomic, strong, readonly) NSMutableSet<NSString *> *linkedServiceNames;
+
+/**
+ This earmarks the user as being an "identity" user.
+ This will make saves write through to the currentUser singleton and disk object
+ */
+@property (nonatomic, assign, getter=isCurrent, setter=setIsCurrent:) BOOL current;
```
richardjrossiii: Because `PFUser` could be subclassed, we should try to pick property names that are less likely to conflict (remember `state`, anyone?). I fear that `lazy` and `current` may already be in use by some applications, and I'd hate to cause incompatibilities for existing code.

nlutsenko: Sure. Let me use the same hack then as we used for '_state' :P

richardjrossiii:
I can't wait for swift...

PR #733, opened 2016-01-06
Update nullability specifier syntax to latest.

nlutsenko:
Since we require Xcode 7 - we can use this syntax now for everything.


richardjrossiii:
Now next beta they're probably going to change it again...

PR #737, opened 2016-01-07
Fix potentially not saved objects if they are not initially dirty.

nlutsenko:
Regression caused by change in #689.
Old code:
```objc
if (!dirty && ![changes count]) {
```
Broken code:
```objc
if (!dirty && change.count != 0) {
```
Fixed code:
```objc
if (!dirty && change.count == 0) {
```

Boolean logic is my weakness.


richardjrossiii:
New rule: Any boolean logic changes shall be looked at in a separate commit and shall be reviewed by no less than 5 members of the team. :smile:

PR #739, opened 2016-01-07
Add Carthage to Travis-CI.

nlutsenko:


richardjrossiii:
Pass back when travis passes, I've re-ran twice no luck.

PR #765, opened 2016-01-16
add includeKeys convenience to PFQuery

AndrewSB:
Thought this was a useful method

### DIFF
--- a/Parse/PFQuery.h
+++ b/Parse/PFQuery.h
@@ -83,7 +83,7 @@ typedef void (^PFQueryArrayResultBlock)(NSArray<PFGenericObject> *_Nullable obje
 ///--------------------------------------
 
 /**
- Make the query include PFObjects that have a reference stored at the provided key.
+ Make the query include `PFObject`s that have a reference stored at the provided key.
 
  This has an effect similar to a join.  You can use dot notation to specify which fields in
  the included object are also fetch.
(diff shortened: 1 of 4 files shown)


nlutsenko:
I think it is, indeed.
Let's do a little bit of work to improve it then...
Pieces that would be required to get this merged in:
- Adding the method to public header file
- Adding documentation on the method in the public header
- Specifying the generic type of the `NSArray *` to be `NSArray<NSString *> *`, since it adds more type safety.
- Adding `includeKeys:` method to `PFMutableQueryState` and using it instead of calling `includeKey:` in a loop, to slightly improve performance (this is nitpicking)

Let me know if you need help with any of these.

AndrewSB:
Sorry about not thinking to do all of that myself 😬.

Haven't done any real Objective-C work in about 18 months. I'll make those changes in the next couple days and ping you when I think it's ready to go 👌

AndrewSB:
@nlutsenko would you like me to also squash all of my commits?

nlutsenko:
Yes, please.

Parse/PFQuery.h
```diff
@@ -95,6 +95,15 @@ typedef void (^PFQueryArrayResultBlock)(NSArray<PFGenericObject> *_Nullable obje
 - (instancetype)includeKey:(NSString *)key;
 
 /**
+ Make the query include PFObjects that have a reference stored at the provided keys.
```
nlutsenko: Nit: `PFObject`s

nlutsenko:
A single nit, otherwise looks great!
Squash - yes, please, if it's not a problem.

Parse/PFQuery.h
```diff
@@ -83,7 +83,7 @@ typedef void (^PFQueryArrayResultBlock)(NSArray<PFGenericObject> *_Nullable obje
 ///--------------------------------------
 
 /**
- Make the query include PFObjects that have a reference stored at the provided key.
+ Make the query include `PFObject`s that have a reference stored at the provided key.
```
AndrewSB: added a "`" to the docs for includeKey as well

nlutsenko: Nice catch, thank you!

PR #790, opened 2016-01-27
Fixes incorrect assertion of 2 week in advance PushDate and adds unit…

jrturton:
… tests to verify


nlutsenko:
Amazing catch. Thank you!

PR #827, opened 2016-02-21
Allow clientKey to be nil, as it's not explicitly required for Open Parse Server.

### LINKED ISSUE #826: [Parse Server] clientKey is optional if using Parse Server
A client key does not need to be configured in order to deploy a Parse Server. The Parse iOS SDK does not currently support connecting to a clientKey-less Parse Server.

To resolve this, a new initialization method is required: 

`+ (void)setApplicationId:(NSString *)applicationId;`

We will also need to remove the assertion that checks for a client key when the SDK is initialized.

nlutsenko:
Implements and closes #826.

cc @hramos

### DIFF
--- a/Parse/Internal/Commands/CommandRunner/PFCommandRunning.h
+++ b/Parse/Internal/Commands/CommandRunner/PFCommandRunning.h
@@ -43,11 +43,11 @@ NS_ASSUME_NONNULL_BEGIN
 
 - (instancetype)initWithDataSource:(id<PFInstallationIdentifierStoreProvider>)dataSource
                      applicationId:(NSString *)applicationId
-                         clientKey:(NSString *)clientKey
+                         clientKey:(nullable NSString *)clientKey
                          serverURL:(NSURL *)serverURL;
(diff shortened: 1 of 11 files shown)


hramos:
LGTM. Any idea why CI failed?

nlutsenko:
Yes, because Carthage comes preinstalled on few machines and we have #828 to handle that.

Parse/Parse.h
```diff
@@ -81,7 +81,7 @@ NS_ASSUME_NONNULL_BEGIN
  @param applicationId The application id of your Parse application.
  @param clientKey The client key of your Parse application.
  */
-+ (void)setApplicationId:(NSString *)applicationId clientKey:(NSString *)clientKey;
++ (void)setApplicationId:(NSString *)applicationId clientKey:(nullable NSString *)clientKey;
```
richardjrossiii: This should still be non null. If the server URL isn't changed (which only happens via the configuration initialization) client key is still required.

PR #859, opened 2016-03-18
Fix Swift 2.2 naming conversion for PFObject.+objectWithoutDataWithClassName:objectId:.

### LINKED ISSUE #806: Swift v2.2: Awkward conversion of ObjC header to Swift ...
From PFObject:

    + (instancetype)objectWithoutDataWithClassName:(NSString *)className objectId:(nullable NSString *)objectId;

Automatically converts to:
Swift v2.1:

    public convenience init(withoutDataWithClassName className: String, objectId: String?)

Swift v2.2:

    public convenience init(outDataWithClassName className: String, objectId: String?)

From PFSubclassing:

    + (instancetype)objectWithoutDataWithObjectId:(nullable NSString *)objectId;

Automatically converts to in both versions but is different than what the PFObject conversion with similar naming:

    optional public static func objectWithoutDataWithObjectId(objectId: String?) -> Self

The different conversions are probably not what was intended. 

Anon,
Andrew

nlutsenko:
Fixes #806.

### DIFF
--- a/Parse/PFObject.h
+++ b/Parse/PFObject.h
@@ -77,7 +77,8 @@ NS_REQUIRES_PROPERTY_DEFINITIONS
 
  @return A `PFObject` instance without data.
  */
-+ (instancetype)objectWithoutDataWithClassName:(NSString *)className objectId:(nullable NSString *)objectId;
++ (instancetype)objectWithoutDataWithClassName:(NSString *)className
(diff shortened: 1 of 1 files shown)


joshuarcher:
what's the difference between "outDataWithObjectId:" and "withoutDataWithClassName: objectId:" ?

nlutsenko:
The former infers the className if it's a subclass of `PFObject`.
The latter requires you to explicitly specify it.

PR #862, opened 2016-03-19
Re-enable Carthage tests.

### LINKED ISSUE #850: Re-enable Carthage on Travis-CI after timeout issue is resolved.
Carthage works great for us when installing locally, but looks like the latest version consistently times out on Travis-CI. This is being discussed in https://github.com/Carthage/Carthage/issues/1163 and we should re-enable it after the issue is resolved.

nlutsenko:
Since it looks like Carthage is fixed - let's try to re-enable it.
Should fix #850.


nlutsenko:
Going to close it to remove the noise. It still fails :(

Issue #625, opened 2015-12-04
Facebook and Twitter login fail if [Parse setApplicationId:clientKey:] is called multiple times


jimspeth:
Our application has a legitimate use for changing Parse apps while running.  In Parse 1.8, this worked without a problem.

However, when we tried upgrading to Parse 1.10, Facebook and Twitter login is failing.  The call to `[PFFacebookUtils logInInBackgroundWithReadPermissions:block:]`is called back with `user` and `error` both nil.  Internally, there is an exception that is caught but not propagated:

```
No registered authentication delegate found for `facebook` authentication type.
Register a delegate first via PFUser.registerAuthenticationDelegate(delegate, forAuthType:)
```

I see in https://github.com/ParsePlatform/Parse-SDK-iOS-OSX/issues/159 that apparently our use case is not supported.  Please consider making this a supported use case and fixing this issue.

nlutsenko:
Hey @jimspeth, unfortunately the behavior is `by design` undefined when calling `Parse.+setApplicationId:clientKey:` multiple times in a single application run.

I will mark this issue as discussion/enhancement, as I can see small narrow set of apps that could benefit from it and we might support it later.

jimspeth:
@nlutsenko How about just exposing a `+[Parse reset]` method that just resets all state?  It seems like this would be fairly straight-forward.

I've been able to work around this in our code by calling these private methods:

```
    [Parse _clearCurrentManager];
    [PFAnonymousUtils _clearAuthenticationProvider];
    [PFFacebookUtils _setAuthenticationProvider:nil];
    [PFTwitterUtils _setAuthenticationProvider:nil];
```

Anything I'm missing resetting?

nlutsenko:
Implemented in #874

PR #874, opened 2016-03-26
Add assertion that makes it invalid to initialize Parse twice in a single runtime.

nlutsenko:
We might want to introduce a method that teardowns Parse completely at a later point as well.
What do you think would be a good name for that method? `resetConfiguration`? `deinitialize`?

Also, needed to cleanup these tests, since the setup/teardown logic was inconsistent.

### DIFF
--- a/Parse/Parse.h
+++ b/Parse/Parse.h
@@ -107,7 +107,7 @@ NS_ASSUME_NONNULL_BEGIN
 /**
  The current client key that was used to configure Parse framework.
  */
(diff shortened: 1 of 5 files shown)


richardjrossiii:
I don't think we should expose an API for deinitializing the SDK at this time. If we get requests for it - maybe... but I doubt the use case is high.

nlutsenko:
We have a request for it - #625.

PR #899, opened 2016-04-14
Add ability to clear PFFile cache.

### LINKED ISSUE #488: PFFileCache does not get cleared out by iOS
We are finding that our application disk usage grows significantly due to the PFFileCache that is stored by Parse.  Our app uploads a log file to the server every 5 min.  There is also another image that is saved numerous times in the cache despite not changing at the client level.  After running for a couple months, the application can take up 8GB of space.

Has a method been provided for flushing this cache manually?  We are not finding iOS to clean it up as Parse says it should.  This is a huge issue for us as our devices run for ~15 hours a day, 7 days a week and they are arriving at a state where the app must be deleted and reinstalled to function again.

nlutsenko:
Adds 2 new methods for clearing cache for a single file or all of them at once.
Fixes #488

### DIFF
--- a/Parse/Internal/File/Controller/PFFileController.h
+++ b/Parse/Internal/File/Controller/PFFileController.h
@@ -94,7 +94,8 @@
 #pragma mark - Cache
 ///--------------------------------------
 
-- (BFTask<PFVoid> *)clearFileCacheAsync;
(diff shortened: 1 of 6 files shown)


richardjrossiii:
LGTM. not sure what the 'more info' tag is for...

nlutsenko:
Sorry, misclick.

PR #928, opened 2016-05-29
Let the iOS SDK handle Delete ops when sent from the server

flovilmart:
Related to: https://github.com/ParsePlatform/parse-server/pull/1946
Original Issue: https://github.com/ParsePlatform/parse-server/issues/1840

In a nutshell:

When performing an unset in Cloud Code the client SDKs need to perform a fetch in order the synchronize their state which is redundant and can possibly break the SoC. 

On parse.com, the bug is present, and the deleted key is not part of the response. No way to know what really happened on cloud code. 

In parse-server, the server responds with a Delete op on the key. 
When serializing the JSON response, `undefined` keys are stripped out of the generated string.
We could have replaced with `null` but that would also require updates on both SDK's to properly handle nulls. and we don't really love `NSNull` on iOS do we?

### DIFF
--- a/Parse/Internal/Object/State/PFObjectState.m
+++ b/Parse/Internal/Object/State/PFObjectState.m
@@ -15,6 +15,7 @@
 #import "PFMutableObjectState.h"
 #import "PFObjectConstants.h"
 #import "PFObjectUtilities.h"
+#import "PFFieldOperation.h"
 
 @implementation PFObjectState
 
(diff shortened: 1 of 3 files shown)


flovilmart:
@nlutsenko any news on that?

ShawnBaek:
@flovilmart Thanks for contribute this repos. Anyway What's going on? @nlutsenko  any update?

PR #947, opened 2016-06-24
objectWithClassName:dictionary: wasn’t encoding the dictionary into PFOject properly

kangaroo5383:
Date type properties are not being converted.

more details here:
https://github.com/ParsePlatform/Parse-SDK-iOS-OSX/issues/945

### DIFF
--- a/Parse/PFObject.m
+++ b/Parse/PFObject.m
@@ -1633,7 +1633,9 @@ static void PFObjectAssertValueIsKindOfValidClass(id object) {
 }
 
 + (instancetype)objectWithClassName:(NSString *)className dictionary:(NSDictionary *)dictionary {
-    return [self _objectFromDictionary:dictionary defaultClassName:className completeData:YES];
+    PFObject *object = [self objectWithClassName:className];
+    [object _mergeAfterSaveWithResult:dictionary decoder:[PFDecoder objectDecoder]];
+    return object;
 }
 
 + (instancetype)objectWithoutDataWithClassName:(NSString *)className objectId:(NSString *)objectId {
--- a/Parse/PFObject.m
+++ b/Parse/PFObject.m
@@ -1634,7 +1634,9 @@ static void PFObjectAssertValueIsKindOfValidClass(id object) {
 
 + (instancetype)objectWithClassName:(NSString *)className dictionary:(NSDictionary *)dictionary {
     PFObject *object = [self objectWithClassName:className];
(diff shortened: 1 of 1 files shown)


facebook-github-bot-9:
Thank you for your pull request and welcome to our community. We require contributors to sign our Contributor License Agreement, and we don't seem to have you on file. In order for us to review and merge your code, please sign up at https://code.facebook.com/cla - and if you have received this in error or have any questions, please drop us a line at cla@fb.com. Thanks!

facebook-github-bot-5:
Thank you for signing our Contributor License Agreement. We can now accept your code for this (and any) Facebook open source project. Thanks!

nlutsenko:
The problem with this approach is that this constructor should mutate the object and keep it `dirty`, whilst using `mergeAfterSaveWithResult:decoder:` will treat the object as fully coming from server and not `dirty`.
The approach that I would recommend is decoding all the keys in the dictionary recursively and then setting the keys instead of going through `mergeAfterSave`.
This will also fix the failing tests.

nlutsenko:
Let me know what you think about using that approach. Requesting changes so it's back into your queue, since it's failing tests...

kangaroo5383:
Thanks @nlutsenko that makes sense, I'll make the change.

Parse/PFObject.m
```diff
@@ -1634,7 +1634,9 @@ + (instancetype)objectWithClassName:(NSString *)className {
 
 + (instancetype)objectWithClassName:(NSString *)className dictionary:(NSDictionary *)dictionary {
     PFObject *object = [self objectWithClassName:className];
-    [object _mergeAfterSaveWithResult:dictionary decoder:[PFDecoder objectDecoder]];
+    [dictionary enumerateKeysAndObjectsUsingBlock:^(id key, id obj, BOOL *stop) {
+        object[key] = [[PFDecoder objectDecoder] decodeObject:obj];
```
nlutsenko: Nit: You can save some method invocations by using a single local variable for storing the decoder. Feel free to leave as is, as it's super nitpicking.

kangaroo5383: +1 nitpicking is important, shouldn't have missed that, my bad.  Didn't think it should be __block since we're not modifying the var.  Thanks for the quick feedback.

nlutsenko: Yup, shouldn't be `__block`, as it doesn't' change by the block.
Looks great nwo.

nlutsenko:
Awesome, this is good to go, just need for all the tests to finish running...

PR #955, opened 2016-07-05
Correctly return a secure URL if https was used as the url of a PFFil…

rdegraci:
This is necessary since Apple will no longer accept App Transport Security setting work-arounds in 2017.

### DIFF
--- a/Parse/Internal/File/State/PFFileState.m
+++ b/Parse/Internal/File/State/PFFileState.m
@@ -78,13 +78,11 @@ static NSString *const _PFFileStateSecureDomain = @"files.parsetfss.com";
     }
 
     NSString *scheme = components.scheme;
-    if (![scheme isEqualToString:@"http"]) {
-        return self.urlString;
+    if ([scheme isEqualToString:@"http"] || [components.host isEqualToString:_PFFileStateSecureDomain]) {
+        components.scheme = @"https";
+        _secureURLString = components.URL.absoluteString;
     }
 
-    if ([components.host isEqualToString:_PFFileStateSecureDomain]) {
-        components.scheme = @"https";
-    }
     _secureURLString = components.URL.absoluteString;
     return _secureURLString;
 }
--- a/Parse/PFFile.m
+++ b/Parse/PFFile.m
@@ -448,6 +448,9 @@
 
 - (NSString *)url {
     __block NSString *url = nil;
+    
+    if (self.state.urlString == nil) return nil;
+    
     [self _performDataAccessBlock:^{
         
         NSURLComponents *components = [NSURLComponents componentsWithString:self.state.urlString];
--- a/Parse/PFFile.m
+++ b/Parse/PFFile.m
@@ -449,7 +449,15 @@
 - (NSString *)url {
     __block NSString *url = nil;
     [self _performDataAccessBlock:^{
-        url = self.state.secureURLString;
+        
+        NSURLComponents *components = [NSURLComponents componentsWithString:self.state.urlString];
+        NSString *scheme = components.scheme;
+        if ([scheme isEqualToString:@"http"]) {
+            url = self.state.urlString;
+        } else {
+            url = self.state.secureURLString;
+        }
+        
     }];
     return url;
 }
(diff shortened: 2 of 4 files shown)


nlutsenko:
Looks like this is actually breaking tests, which means that it's breaking the functionality.
We support ATS without any workarounds and files uploaded recently to Parse.com (within last couple of years) or files that have `https` scheme in the server-returned URL will be downloaded using SSL turned on.

Are you running into any specific problems? Would love to know what is your use case and why you think this change is required.

Also, please note that in order to get the secure url from a file - you should use `secureUrlString` instead of `urlString` property.

rdegraci:
Sorry, I had forgotten that there were specific OS X tests and I had not run **rake test:osx**  before submitting the Pull Request.  I've added another commit fix to make the OS X tests pass.

nlutsenko:
Are you running into any specific problems? Would love to know what is your use case and why you think this change is required.

Parse/Internal/File/State/PFFileState.m
```diff
@@ -78,13 +78,11 @@ - (NSString *)secureURLString {
     }
 
     NSString *scheme = components.scheme;
-    if (![scheme isEqualToString:@"http"]) {
-        return self.urlString;
-    }
-
-    if ([components.host isEqualToString:_PFFileStateSecureDomain]) {
+    if ([scheme isEqualToString:@"http"] || [components.host isEqualToString:_PFFileStateSecureDomain]) {
```
nlutsenko: You are forcing everything here to use `https`, even though the url being used was `http`. The intention of this code is to use `https` if we can (either we are using a parse.com file domain or we aren't using `http` sent from the server.

nlutsenko:
This sounds right, but from what I can see (and I might be wrong here) - the current test (the one included here) doesn't fail on the current latest `master`.
Can you provide a test that is failing?

nlutsenko:
I also tested the file download/upload with the current state of master and a file that was created using `https` URL - everything is using `https` urls as long as the server is sending an `https` in the `url` field of a file (which is the intended behavior).

rdegraci:
I added a test which passes within this Pull Request, but fails in the master branch. The test shows that when using the AWS S3 host (it is configured when the S3 Adapter is installed), the secureUrlString has the incorrect "http" protocol, when it should have "https".

nlutsenko:
Turns out that it actually has the correct host. Unless it's parse.com S3 files bucket - we can't guarantee that `https` is the scheme that we can use, which means that if people want to opt-out of `https` scheme -we still are forcing them into it.

I would recommend opening a PR against S3Adapter to change the configuration there (add configuration option on whether `https` should be used at all times) instead of here, as it's clearly a major breaking change and we would want to avoid it.

Parse/Internal/File/State/PFFileState.m
rdegraci: I am confused and I think some clarification:

The name of the function is secureURLString, should it always return the "https" version of the self.urlString ?

nlutsenko: Since this is internal API, it does indeed need some clarification from our end:

- `secureURLString` is the one that we are trying to use at all times by default
- `urlString` is the server-provided default URL
- Since we only know that we can automatically update `http` to `https` for parse.com S3 bucket - there is a hard-coded check on that domain
- For every other domain - whatever is sent from the server is the source of truth and we are not going to change it, since the server dictates where the files should be downloaded from.

nlutsenko:
Let me also give you a concrete example of when people can encounter this and it would be a wrong choice of logic:

- You are storing Parse Files in a custom location and are not using `ssl` to protect it, meaning that you are sending the location of files back to the clients with `http` scheme.
- ObjC SDK is going to automatically update any `http` scheme to `https` if this change is merged
- Since there is no support for `https` on the stored location - the client is going to fail to download any file
- Panic & Freak Out

Once again, I believe the change to serve files over `https` from S3 should go directly into the S3Adapter and I am more than happy to review any Pull Requests that change or add configuration for that logic on Parse Server / Parse Server Modules.

Issue #963, opened 2016-07-07
PFUser Subclass currentUser() not returning subclass


Vortec4800:
Say we have a custom user subclass:

```swift
class TUser: PFUser {
	override class func initialize() {
		struct Static {
			static var onceToken : dispatch_once_t = 0;
		}
		dispatch_once(&Static.onceToken) {
			self.registerSubclass()
		}
	}
	
	// MARK: Parse Properties
	@NSManaged var mySpecialObject: PFObject?
}
```

The expectation is that TUser.currentUser() will return the currently authenticated user as a TUser object. Unfortunately, we often get a PFUser object back instead, and the method ignores our subclass altogether.

When trying to access a property like `TUser.currentUser()?.mySpecialObject` this causes a bad access crash, because PFObject does not have a mySpecialObject property.

I tried to do a little digging and got all the way to the `PFObjectFilePersistenceController` and the `PFObjectFileCoder` where it appears to write the current user info to Core Data, and it stores the Parse class name (_User in our case) but doesn't seem to do anything for subclassing. Following the call tree up, I don't see anywhere that a subclass would get instantiated instead and returned.

How is this supposed to work? Is the issue stemming from creating these subclasses using Swift? Is the Parse SDK expecting something dynamic that Swift isn't providing?

Vortec4800:
Some more info. It seems like the Parse initialize method pre-loads "current" objects from disk. This seems to happen before the custom `TUser` subclass is registered. The current user controller then caches the current user, which is a `PFUser` object, and that's what gets vended from there on out.

Running the `TUser.registerSubclass()` before calling Parse.init should solve this problem. Going to keep looking to see if there is another solution that doesn't require moving the register subclass method.

Vortec4800:
Created a PR which will fix the issue without any code changes. This should result in safer code (doesn't matter if some developers are unaware of this bug or that they may need to move their subclass reg methods, if they update their code will just work as expected.

nlutsenko:
@Vortec4800, good job on identifying the root cause of this.
Since currently it's required to register subclasses before Parse is initialized - I am not sure if the temporary workaround is required at all and I would rather pursue #5 (as you said) instead of adding a temporary patch around something that will go away.

Let me know your thoughts on this, so we can discuss and make the best decision together.

Vortec4800:
I had a few thoughts about that.

[...]

PR #964, opened 2016-07-07
Added object class check to current user controller's cache

Vortec4800:
This is a fix for the issue identified in #963.

If the `_currentUser` object is set early on before a custom user subclass has been registered, it will never be updated with the subclassed user. As the current user loading is asynchronous this can create a race condition, and for some app launches `currentUser()` will return a PFUser and other times it will return your custom subclass.

This adds a check to verify that the cached user object is of the class that we expect to get. If it isn't, it continues with user loading which will create the correct class.

This all means that `currentUser()` will return a PFUser before the user subclass is registered, and a `MYUser` custom subclass after the registration completes. This seems like expected behavior and should prevent hard to track down bugs, especially in Swift projects where the + load method isn't available.

As an aside, it seems this issue may also be fixed by the merging of #5. That has been sitting around for a long time, and this PR should get us by until that old PR can be evaluated and merged.

### DIFF
--- a/Parse/Internal/User/CurrentUserController/PFCurrentUserController.m
+++ b/Parse/Internal/User/CurrentUserController/PFCurrentUserController.m
@@ -106,7 +106,7 @@
 		if (currentUser) {
 			// Verify that the current user matches our expected registered subclass
 			Class expectedClass = [[PFObjectSubclassingController defaultController] subclassForParseClassName:@"_User"];
-			if(expectedClass == [currentUser class]){
+			if([currentUser isKindOfClass:expectedClass]){
 				return currentUser;
 			} else {
 				matchesDisk = NO;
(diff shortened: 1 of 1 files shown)


Parse/Internal/User/CurrentUserController/PFCurrentUserController.m
```diff
@@ -102,9 +103,16 @@ - (BFTask *)_getCurrentUserAsyncWithOptions:(PFCurrentUserLoadingOptions)options
             matchesDisk = _currentUserMatchesDisk;
             currentUser = _currentUser;
         });
-        if (currentUser) {
-            return currentUser;
-        }
+		if (currentUser) {
+			// Verify that the current user matches our expected registered subclass
+			Class expectedClass = [[PFObjectSubclassingController defaultController] subclassForParseClassName:@"_User"];
+			if(expectedClass == [currentUser class]){
```
nlutsenko: We probably should do `[currentUser isKindOfClass:expectedClass]` here instead of direct class comparison.

Vortec4800:
Probably a good idea. I've been working in Swift too much lately. :D

nlutsenko:
Let's continue the discussion on the linked issue. This looks good and I love the idea, but not sure whether a temporary patch is the right solution.

Vortec4800:
Updated as requested, in case we decide to use this.

PR #1029, opened 2016-09-10
Made PFEncoder/PFDecoder public

agordeev:
As @richardjrossiii [suggested](https://github.com/ParsePlatform/ParseLiveQuery-iOS-OSX/issues/64) I've made `PFEncoder`/`PFDecoder` public, so Parse LiveQuery could use it instead of it's own parsing implementation.


flovilmart:
@agordeev can we get that in finally?

flovilmart:
@agordeev I'll take it over, there are some issue with the headers not being properly marked public in the targets.

dplewis:
+1

PR #1057, opened 2016-10-18
Remove automatic registration of subclasses

zadr:
This effectively reverts e9a4e59 to remove automatic subclass registration.

For unknown reasons, this is broken on iOS 10 for a large number of people. Could spend the time debugging this, but, could also keep on keeping on with existing code that uses manual registration that isn't crashing ¯\_(ツ)_/¯


zadr:
oops, i didn't mean to send this upstream, sorry.

PR #1069, opened 2016-11-09
Load bundle before resgistering subclasses

jmfieldman:
Intended to address https://github.com/ParsePlatform/Parse-SDK-iOS-OSX/issues/1006


jmfieldman:
Closing this because of reports that it does not solve the problem.

PR #1087, opened 2017-01-04
Prevent crash by loading the bundle if it's not loaded yet (see: http…

jordanhaven:
…s://github.com/ParsePlatform/Parse-SDK-iOS-OSX/issues/1006 ):

This is the fix from @funnel20. Just opening a PR to speed things along.


levigroker:
It's a nitpick, but `NSError *error = nil;` should be `__autorelease NSError *error = nil;`, IMHO

PR #1090, opened 2017-01-10
Safely inspect bundles for classes

### LINKED ISSUE #1006: 1.14.2: "Cannot register subclasses in a bundle that hasn't been loaded!"
Just to report a crashlytics event that has occurred for a few users on a build that uses the Parse 1.14.2 SDK.  It's crashing in `[PFObjectSubclassingController _registerSubclassesInBundle:]` at the assertion that bundle.loaded is true.  I can see from the report that it is being called from the NSBundleDidLoadNotification notification handler.  There must be some inconsistency between that handler being called and the `loaded` property of the bundle being passed into the notification.

flovilmart:
Fixes #1006

Makes sure we don't check classes in not loaded bundles.

### DIFF
--- a/Parse/Internal/Object/Subclassing/PFObjectSubclassingController.m
+++ b/Parse/Internal/Object/Subclassing/PFObjectSubclassingController.m
@@ -128,20 +128,12 @@ static NSNumber *PFNumberCreateSafe(const char *typeEncoding, const void *bytes)
                                                                                             queue:nil
                                                                                        usingBlock:^(NSNotification *note) {
                                                                                            @strongify(self);
-                                                                                           [self _registerSubclassesInBundle:note.object];
+                                                                                           [self _registerSubclassesInLoadedBundle:note.object];
                                                                                        }];
     }
     NSArray *bundles = [[NSBundle allFrameworks] arrayByAddingObjectsFromArray:[NSBundle allBundles]];
     for (NSBundle *bundle in bundles) {
-        // Skip bundles that aren't loaded yet.
-        if (!bundle.loaded || !bundle.executablePath) {
-            continue;
-        }
-        // Filter out any system bundles
-        if ([bundle.bundlePath hasPrefix:@"/System/"] || [bundle.bundlePath hasPrefix:@"/Library/"]) {
-            continue;
-        }
-        [self _registerSubclassesInBundle:bundle];
+        [self _registerSubclassesInLoadedBundle:bundle];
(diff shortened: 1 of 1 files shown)


flovilmart:
@hramos @nlutsenko, we are kinda stuck here with that one. Could you review, advise?

flovilmart:
Thanks @hramos !

partikles:
@flovilmart @hramos . Our startup almost completed our project using ParseSever with iOS&Android SDK. We planning to go live.  Do we expect iOS/Android SDK get the update like facebook does? We need to inform this to investors. Very critical. 

Please guide us.

flovilmart:
@hramos I'll need the commit access to the repo as Cocoapods require the tag to be pushed before 
I can send pod to trunk. Opening #1111

hramos:
@flovilmart invited

flovilmart:
Turns out I don't have access to the pods either ;/ 

```
!] You (florent@flovilmart.com) are not allowed to push new versions for this pod. The owners of this pod are nlutsenko@me.com, grantland@fb.com, hramos@fb.com, and cocoapods@fb.com.
```

levigroker:
Anyone have pod access to push out 1.14.3? Please?

levigroker:
@flovilmart We manually put this change in our now-production app and are seeing a crash (unknown selector sent to instance). Specifically, it appears that `_registerSubclassesInLoadedBundle:` can be called with objects (NSString apparently) which are not NSBundle objects, and the `loaded` message is sent to an object which doesn't handle the selector. I'd guess this is coming from the `NSBundleDidLoadNotification` `note.object` but haven't verified that.

An obvious armor for this is to ensure the object is of the intended type in `_registerSubclassesInLoadedBundle`.

i.e. Change:

    if (!bundle.loaded || !bundle.executablePath) {

to

    if (![bundle isKindOfClass:NSBundle.class] || !bundle.loaded || !bundle.executablePath) {

(though it would be nice to understand why we're not getting NSBundle objects, this will prevent the crash).

Since 1.14.3 has yet to be released, this change (or similar) prior to release would be appreciated.

Cheers,

Levi

flovilmart:
The changes have been merged to master, can you open a Pr with your suggestion and the fix?

levigroker:
@flovilmart Cheers. Here's the pull: https://github.com/ParsePlatform/Parse-SDK-iOS-OSX/pull/1119

PR #1102, opened 2017-02-03
Update documentation links

jaimeagudo:
They were hard to find due redirect loop
Make explicit the deprecation of push notifications on parse-server


jaimeagudo:
Following directions from this issue https://github.com/ParsePlatform/parse-server/issues/3455. Perhaps we need to make explicit the intended use of master-key as per the thread

flovilmart:
Thanks for the PR! Merging now!

PR #1112, opened 2017-02-20
Added a property to control keychain identifiers

laptobbe:
I ran into a problem with using parse across multiple apps with app groups. The implementation only took into account sharing keychain access to extensions. Causing saves on the current user to fail.

My solution was to add a property so that you can specify exactly what keychain identifier will be used. I don't know if this is the best solution, I'am open for suggestions :)


flovilmart:
@laptobbe why don't you use the : `containingApplicationBundleIdentifier` does it have other side effects?

laptobbe:
@flovilmart I tried that initially, however it is only allowed in an extension environment. I did not look further into why this is, I just worked around it. See the code below.

``` objc
+ (void)initializeWithConfiguration:(ParseClientConfiguration *)configuration {
    PFConsistencyAssert(configuration.applicationId.length != 0,
                        @"You must set your configuration's `applicationId` before calling %s!", __PRETTY_FUNCTION__);
    PFConsistencyAssert(![PFApplication currentApplication].extensionEnvironment ||
                        configuration.applicationGroupIdentifier == nil ||
                        configuration.containingApplicationBundleIdentifier != nil,
                        @"'containingApplicationBundleIdentifier' must be non-nil in extension environment");
...
```

flovilmart:
I'm not 100% familiar with keychain sharing, but it seems that lifting the restriction on requiring the extension environment would be a cleaner solution as in both scenarios, you want sharing a keychain across multiple bundle identifiers right? Or am I mistaken?

laptobbe:
@flovilmart You are right that is what I am trying to achieve.

I agree with you it's a cleaner solution as long as there are no other side-effects when we lift the restriction. I can look into it and see if there is anything else we need to consider.

flovilmart:
Awesome!! Let's investigate that, maybe the restriction should be the other way around, if in extension environment, we require the option to be set.

flovilmart:
Actually it's not the 'other way around' as the message is clear, i don't believe the restriction should apply setting that option when not in extension environment, it looks like an overlooked use case.

laptobbe:
@flovilmart I might be wrong here, I am not super familiar with the code.

But at first glance it seems that setting `containingApplicationBundleIdentifier` for a non extension would break the application group container sharing. Because `containingApplicationBundleIdentifier` is assumed to only be set for an extension and therefore the data is not migrated to the shared container. However I think we can change this easily to check the  environment property instead.

Here is the code that handles the migration.

```objc
- (void)_migrateSandboxDataToApplicationGroupContainerIfNeeded {
    // There is no need to migrate anything on OSX, since we are using globally available folder.
#if TARGET_OS_IOS
    // Do nothing if there is no application group container or containing application is specified.
    if (!self.configuration.applicationGroupIdentifier || self.configuration.containingApplicationBundleIdentifier) {
        return;
    }
   ...
}
```

flovilmart:
> However I think we can change this easily to check the environment property instead.

That seems to be reasonable.

laptobbe:
@flovilmart I did the update to check for the environment, and it seems to work in my project.

I could not get the tests to run locally due to build errors I didn't have time to look into. There might be some tests with regard to the change that needs to be updated.

flovilmart:
We'll see how it goes through CI

flovilmart:
@laptobbe sorry for the late response. Kinda dropped the ball on the SDK. Seems that the CI is not really happy with the changes. Can you investigate?

PR #1130, opened 2017-04-06
Fixed issues where a nil parameter would cause a crash, added nullability specifiers

charliewilliams:


flovilmart:
Thanks for the PR, among all of the changes I have trouble identifying where the issue is. We're just merged the support for swift 3, so you should be probably fine rebasing and taking if from there. Can you please reopen a PR with just the fix and an appropriate description or a link to the issue?

charliewilliams:
Reopened (re-created, really) as https://github.com/parse-community/Parse-SDK-iOS-OSX/pull/1187 and https://github.com/parse-community/Parse-SDK-iOS-OSX/pull/1188

PR #1141, opened 2017-05-15
Removed or replaced the broken links in the README

davidolesch:
1. I removed the parse.com link because it redirects to https://parseplatform.github.io which is currently a 404 page.
2. I changed the docs link to the parseplatform.org domain.
3. I changed the blog link from https to http because https doesn't load.


flovilmart:
Thanks for tidying this up!

PR #1144, opened 2017-06-05
Fix PFCommandCache memory leak

antoinelamy:
screen shot 2017-06-05 at 8 13 00 am

### DIFF
--- a/Parse/Internal/PFEventuallyQueue.m
+++ b/Parse/Internal/PFEventuallyQueue.m
@@ -174,7 +174,9 @@ NSTimeInterval const PFEventuallyQueueDefaultTimeoutRetryInterval = 600.0f;
 ///--------------------------------------
 
 - (void)start {
+    @weakify(self)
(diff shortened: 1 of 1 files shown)


flovilmart:
Nice Thanks!

antoinelamy:
The builds seems stuck on Travis, is this a known issue? I doubt that this commit is causing this.

flovilmart:
Did you try running the script that is run on CI locally?

antoinelamy:
It took me a while to recreate the travis environment but I managed to get the script working and it is confirmed that my modification cause a regression. It's a bit odd but it seems that the `PFEventuallyQueue` is responsible for retaining itself by design. It will take more time to investigate so I'm closing this PR for the moment.

flovilmart:
Also, a cycle is not necessarily a leak. Are those objects leaked when deallocated?

antoinelamy:
@flovilmart in my case it was a leak

PR #1150, opened 2017-06-20
Add withinPolygon to Query

dplewis:
https://github.com/parse-community/parse-server/pull/3866


flovilmart:
Now that 2.5.0 is released that's great!

PR #1157, opened 2017-07-11
Add Polygon Type and PolygonContain to Query

dplewis:
https://github.com/parse-community/parse-server/pull/3944

I've also added a containsPoint function that can be used for live queries.


dplewis:
@flovilmart can you look at this

PR #1164, opened 2017-07-26
Fix #1162 - PFSubclassing.object() class method unavailable since XCode 9 beta 4

### LINKED ISSUE #1162: XCode 9: PFSubclassing.object() breaks builds on XCode 9 beta 4
Not sure what exactly is happening, but XCode 9 beta 3 started emitting this as a warning message:

```Unavailable class method 'object()' was used to satisfy a requirement of protocol 'PFSubclassing'```

and XCode 9 beta 4 marked this as an error making it impossible to even build the project.

Since ``PFSublcassing.object()`` is an optional requirement, commenting out the method in the ``PFSubclassing.h`` header files makes the project build again - assuming you do not rely on this functionality.

Could be a bug in the XCode 9 beta 4, could be an abuse of ``object()`` method in Parse iOS SDK.

Any thoughts how to fix this?

### PR DESCRIPTION
**Pull request by mman**
The fix essentailly renames the original ``PFSubclassing.object()`` class method to ``PFSubclassing.instantiateObject()`` method with the same semantics to avoid being flagged by XCode 9 beta 4.

### DIFF
--- a/Parse/PFSubclassing.h
+++ b/Parse/PFSubclassing.h
@@ -33,18 +33,6 @@ NS_ASSUME_NONNULL_BEGIN
 @optional
 
 /**
- Constructs an object of the most specific class known to implement `+parseClassName`.
-
- This method takes care to help `PFObject` subclasses be subclassed themselves.
- For example, `PFUser.+object` returns a `PFUser` by default but will return an
- object of a registered subclass instead if one is known.
- A default implementation is provided by `PFObject` which should always be sufficient.
-
- @return Returns the object that is instantiated.
- */
-+ (instancetype)object;
-
-/**
  Creates a reference to an existing PFObject for use in creating associations between PFObjects.
 
  Calling `PFObject.dataAvailable` on this object will return `NO`
--- a/Parse/Internal/Object/PFObjectPrivate.h
+++ b/Parse/Internal/Object/PFObjectPrivate.h
@@ -140,6 +140,22 @@
 
 @end
 
+@interface PFObject ()
+
+/**
+ Constructs an object of the most specific class known to implement `+parseClassName`.
+
+ This method takes care to help `PFObject` subclasses be subclassed themselves.
+ For example, `PFUser.+object` returns a `PFUser` by default but will return an
+ object of a registered subclass instead if one is known.
+ A default implementation is provided by `PFObject` which should always be sufficient.
+
+ @return Returns the object that is instantiated.
+ */
++ (instancetype)object;
+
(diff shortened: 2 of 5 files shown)

### CODE REVIEW

File: Parse/PFSubclassing.h [hunk 1]
```diff
@@ -36,13 +36,13 @@ NS_ASSUME_NONNULL_BEGIN
...
  This method takes care to help `PFObject` subclasses be subclassed themselves.
- For example, `PFUser.+object` returns a `PFUser` by default but will return an
+ For example, `PFUser.+instantiateObject` returns a `PFUser` by default but will return an
  object of a registered subclass instead if one is known.
  A default implementation is provided by `PFObject` which should always be sufficient.
 
  @return Returns the object that is instantiated.
  */
-+ (instancetype)object;
++ (instancetype)instantiateObject;
```
flovilmart: actually, ```A default implementation is provided by `PFObject` which should always be sufficient.``` So it should be good to remove in PFSubclassing altogether no?

mman: I tried removing it but it's actually used by internal PFInstallation controller somewhere so it will need to be moved to either a private header or a private interface of PFInstallation at least.

I do not use it myself and don't have enough experience to assess whether it's safe to remove it from the interface, but from what I could gather studying the docs the point was to register custom subclasses of PFInstallation and PFUser before initializing Parse API and getting the custom subclasses returned from the SDK, although I found it only being used for PFInstallation in fact...

flovilmart: so let's make it private no? it's handled in PFObject.h right but not exposed publicly*

mman: Makes sense, let me jump on it and retest it properly tomorrow...

mman:
Moved the class method PFSubclassing.object to a private category interface on PFObject, not sure if its clean but compiles and works.

flovilmart:
Seems that the tests are also using it, id the method exposed on PFObject publicly?

mman:
I've moved it to PFObjectPrivate.h which is never exposed to swift. From what I could see the moment +object() is exposed via public header to swift the compiler will start complaining. Solution 1: include PFObjectPrivate in tests. Solution 2: Keep the methods public in PFSubclassing.h and rename it everywhere. Ideas?

flovilmart:
go for solution 1 :)

mman:
solution 1 implemented, tests passing on my machine. /me learning slowly :)

flovilmart:
@mman better than not trying at all! 👏 👏 🙇

PR #1187, opened 2017-09-12
Fix some issues where nil was passed to a method requiring a non-nil parameter

charliewilliams:
These are all crashes in the client, all require a non-nil value.


charliewilliams:
I agree that nullability would be the ideal change here, but 1) these are internal headers, and 2) there aren't any nullability annotations currently, so it's a nontrivial amount of work to figure out what the nullability status should be for the entire header file in these cases.

Method params like `selectedKeys` in `PFObject:779` come from a `PFQueryState` object like so:

`NSArray *selectedKeys = queryState.selectedKeys.allObjects;`

So if we enforce a non-nil value in the method I've changed, we're just requiring the caller to do a nil check— there's not a way to require `queryState.selectedKeys.allObjects` to not be nil.

flovilmart:
@charliewilliams let's merge it and release!

PR #1188, opened 2017-09-12
Remove PFEventuallyQueueTestHelper because (1) it was causing memory leaks in production code

charliewilliams:
…and (2) it wasn't actually used in any tests.


flovilmart:
Can’t we just use macros so the test helper is included only in test builds?

charliewilliams:
My initial thought was to only include the test helper on test builds. It doesn't seem to actually be used in any tests, though, so having it in doesn't seem to confer any benefit

flovilmart:
Seems that the build is passing correctly, so let's go with it!

PR #1219, opened 2017-12-04
Automated releases

flovilmart:
- Adds automated Github releases upload on tags
- Adds automated Cocoapods releases on tags
- Adds automated docs generation through Jazzy on master merges (gh-pages force push)
- Revamps Travis.yaml to leverage build stages
- Generated docs demo available here: http://flovilmart.github.io/Parse-SDK-iOS-OSX/api/


flovilmart:
Next on the bucket list

- Release docs for all other projects in their own GH-pages (stop aggregating everything)
- Update the docs.parseplatform.org / readmes to point to specific repos's docs (which makes more sense and easier to maintain)

montymxb:
Seems the release stage is hanging the CI?

flovilmart:
It’s just taking forever as Travis only allows 2 concurrent Mac based builds now :/

flovilmart:
Sorry for the zillion commits :) it gives a sense of how s****y it is to configure it :)

flovilmart:
@montymxb just finished :)

.travis.yml
```diff
@@ -1,38 +1,70 @@
...
-  matrix:
-    - TEST_TYPE=iOS
-    - TEST_TYPE=macOS
-    - TEST_TYPE=Deployment
-    - TEST_TYPE=Carthage
-    - TEST_TYPE=CocoaPods
+branches:
+  only:
+    - master
+    - /^v?[0-9]+.[0-9]+.[0-9]+(-.*)?$/
```
montymxb: You're going to want to escape those `.`s. In a regex it's interpreted as a special 'match any' character unless it's escaped by a preceding `\`. This regex could match stuff like `123` or `1a2b3c` in addition to `v1.2.3`. The exception being if it's in a character class, such as `[.]`, where it's interpreted just as it is.

flovilmart: uhm that's a good point! I need to update it all over the place then :)

README.md
```diff
@@ -65,6 +65,18 @@ Notice the API docs aren't totally up to date when it comes to latest Swift sign
 
 We want to make contributing to this project as easy and transparent as possible. Please refer to the [Contribution Guidelines][contributing].
 
+## Preparing for a new release
```
montymxb: I'm not sure if we want this in the README. Although it is helpful for us it's not something we necessarily need to show to users. This might be more appropritae in CONTRIBUTING I think.

flovilmart: Yeah, many project have it in the README, it's worth having in case you forget in any case, those are part of the scripts. The bumping just changes the versions all over the required files, nothing dangerous in there. I'd keep it there.

montymxb: Alright, fair enough. That does make sense if we want new individuals (or existing) to be able to follow a standard for releasing a new version.

flovilmart: I found myself multiple times wondering how to bump the version number :)

PR #1230, opened 2018-01-14
Merge all ObjC Frameworks

flovilmart:
This PR integrates the FacebookUtils and TwitterUtils repo into a single codebase.
We'll probably need to re-work further the readme.

The motivation is to keep a single codebase for those libraries, and be able to distribute them all at once, instead of managing a lot of boilerplate ci code.

It also uses circle CI for the build status, and travis for releases deployment (easier to push to github for now etc...)

What's to be reviewed? Probably just the README


README.md
```diff
@@ -27,12 +28,30 @@ Notice the API docs aren't totally up to date when it comes to latest Swift sign
...
+  If you wish to use the Facebook or Twitter utils or ParseUI,
+  you can now leverage Cocoapods 'subspecs'
+
+  ```ruby
+    pod 'Parse/FacebookUtils'
+    pod 'Parse/TwitterUtils'
+    pod 'Parse/UI
+  ```
+
+  Note that in this case, the Parse framework will contain all header, so you just have to use:
```
montymxb: typo
> will contain all header**s**

README.md
```diff
@@ -107,12 +135,9 @@ of patent rights can be found in the PATENTS file in the same directory.
 As of April 5, 2017, Parse, LLC has transferred this code to the parse-community organization, and will no longer be contributing to or distributing this code.
 
  [docs]: http://docs.parseplatform.org/ios/guide/
- [blog]: http://blog.parse.com/
```
montymxb: No blog link?

flovilmart: I removed it as it may not contain any relevant information anymore about this repo. This seemed to be an artifact of the old parse.com where some articles would be pushed out about new SDK features or demo usages.

flovilmart:
@montymxb I copy pasted the code from the other repos.

flovilmart:
and addressed the nits ;)

README.md
montymxb: That's fine then. I want to get back to a point where we consistently put up posts about what we're doing, like coalescing all the ObjC frameworks 😉 .

flovilmart: yeah, I'll write something this week, when deprecating / archiving the other repos.

PR #1236, opened 2018-01-19
Set bolts back to 1.8.4

flovilmart:
Until we have a proper solution for not throwing exceptions within tasks (error passing across the SDK?) We have to properly continue to function without throwing errors.

This reverts bolts to 1.8.4, which still has the task exception handling.

### DIFF
--- a/Cartfile
+++ b/Cartfile
@@ -1,1 +1,1 @@
-github "BoltsFramework/Bolts-ObjC" ~> 1.9
+github "BoltsFramework/Bolts-ObjC" == 1.8.4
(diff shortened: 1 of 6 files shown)


montymxb:
@flovilmart do we want to revert first and then work on #1237, or focus on just that (it is the fix for this right?).

flovilmart:
Yeah, i’m Not sure, i’d rather not revert as bolts has a breaking Swift API :/ let’s see how far I can go

montymxb:
Gotcha, let's hold until we see whether you can work it out in the other one.

Issue #1202, opened 2017-10-26
Error: Attempted to change an objectId to one that's already known to the Offline Store


StratRob:
Hi everyone,

I'm having a regular issue with my project, using Parse Server. First, I call a Parse Cloud function to populate a user's data list :

```
var dataSet: Set<MyData>?

func loadData(withParameters parameters: [String : Any]) {

        PFCloud.callFunction(inBackground: "loadData", withParameters: parameters) { (success, error) in

            if let objects = success as? [[String : Any]] {

                let dataTable: [MyData] = objects.map({ (object) -> MyData in

                    let myData = MyData(dataSource: PFObject(className: "MyData", 
                                        dictionary: object))
                    myData.dataSource?.objectId = object["objectId"] as? String
                    return myData
                })

                if self.dataSet == nil {
                    self.dataSet = []
                }

                self.dataSet = Set(dataTable)
            }
      }
} 
```

On the code mentioned above, I have to set the objectId because without this, no matter how many objects I fetch from Parse, when I reduce the array to a set with the last instruction I end up with only one object in it.

However, although this works, when I call this function again to update user's data, I get this error on the `myData.dataSource?.objectId = temp["objectId"] as? String` line : 

> Attempted to change an objectId to one that's already known to the Offline Store.

Any idea? Thanks for your help.

iThinkersTeam:
Have same problem :(

flovilmart:
@StratRob when calling a cloud function, the objects should already by hydrated as PFObject’s. Can you confirm that they are indeed not dictionaries? PFObject can be casted as dictionaries as they also conform to subscripting extensions.

ceramicatheist:
I can confirm we got a new flood of this exact crash when deploying a new version of our app that upgraded from 1.15.3 to 1.15.4, with the associated Bolts upgrade, with no app changes that should have triggered something like this.
(apologies for the long stack trace...)
```
Attempted to change an objectId to one that's already known to the OfflineStore

Fatal Exception: NSInternalInconsistencyException
0  CoreFoundation                 0x181306364 __exceptionPreprocess
1  libobjc.A.dylib                0x18054c528 objc_exception_throw
2  CoreFoundation                 0x1813062ac -[NSException initWithCoder:]
3  Hidrate                        0x100966d20 -[PFOfflineStore updateObjectIdForObject:oldObjectId:newObjectId:] (PFOfflineStore.m:1027)
[...]

PR #1256, opened 2018-02-06
Hot fix for #1237 and #1202

ananfang:
~~1. Fix typo PFPreconditionFailOnError and PFPreconditionFailAndSetError.~~

2. If there is a new object in OfflineStore, should not send the exception when the saved object update its OfflineStore cache:

change the assert condition in function `- (void)updateObjectIdForObject:oldObjectId:newObjectId:` of PFOfflineStore.m

```
// If this object is just saved, update local store without exception.
BOOL isNewlySaved = (oldObjectId == nil) && (object.createdAt != nil);
PFConsistencyAssert(isNewlySaved || existing == nil || existing == object,
                    @"Attempted to change an objectId to one that's already known to the OfflineStore. className: %@ old: %@, new: %@",
                    className, oldObjectId, newObjectId);
```

### DIFF
--- a/Parse/Parse/Internal/Commands/PFRESTQueryCommand.m
+++ b/Parse/Parse/Internal/Commands/PFRESTQueryCommand.m
@@ -24,7 +24,7 @@
 
 + (nullable instancetype)findCommandForQueryState:(PFQueryState *)queryState withSessionToken:(NSString *)sessionToken error:(NSError **)error {
     NSDictionary *parameters = [self findCommandParametersForQueryState:queryState error:error];
-    PFPreconditionBailOnError(parameters, error, nil);
+    PFPreconditionFailOnError(parameters, error, nil);
     return [self _findCommandForClassWithName:queryState.parseClassName
                                    parameters:parameters
                                  sessionToken:sessionToken
@@ -51,7 +51,7 @@
                                                        extraOptions:extraOptions
                                                      tracingEnabled:trace
                                                               error:error];
-    PFPreconditionBailOnError(parameters, error, nil);
+    PFPreconditionFailOnError(parameters, error, nil);
     return [self _findCommandForClassWithName:className
                                    parameters:parameters
                                  sessionToken:sessionToken
@@ -68,7 +68,7 @@
                                                  parameters:parameters
                                                sessionToken:sessionToken
                                                       error:error];
-    PFPreconditionBailOnError(command, error, nil);
+    PFPreconditionFailOnError(command, error, nil);
     return command;
 }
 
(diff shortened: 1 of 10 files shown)


flovilmart:
@ananfang thanks for the PR, PFPreconditionFailOnError doesn’t exist I’m not even sure how it compiles, can you revert to PFPreconditionFailOnError? This seem unrelated to that PR.

ananfang:
@flovilmart 
I just reverted PFPreconditionBailOnError and PFPreconditionBailAndSetError.

ananfang:
@flovilmart 
Thanks for your help 👍 

This hotfix is just a workaround, the actual problem is SDK assigns twice (or more) objectId to newly saved object. The first time objectId is assigned, `existing == nil`, pass the assert. But the second time, `existing != nil && existing != object`, raises the exception.

flovilmart:
DO you have some code that reproduces the issue?

ananfang:
Followings are my scenario:
1) A chat view with two users.
2) Use ParseLiveQuery to listen messages created.
3) User sends message as a PFObject, save it by saveInBackground() method.

It will not always crash, about 50/50.

flovilmart:
Are you saving/mutating the received messages from liveQuery?

ananfang:
No, I don't mutate messages from LiveQuery

flovilmart:
Would you be able to share a demo project that reproduces the issue? I believe it’s related to liveQuery

brianyyz:
+1

ananfang:
@flovilmart 
[Demo Project](https://github.com/ananfang/parse-testing-app)

Run and tap "Add Message" button, it will crash all the time by "Attempted to change an objectId to one that's already known to the Offline Store"

ananfang:
I found a PFObject Equality bug comes from my "isNewlySaved" workaround.

Following is scenario:
1. Enable Local Datastore
2. Use Live Query to listen new Message object (PFObject) created.
3. Save a new Message object.
4. There will be two Message objects assigned the same new objectId.
5. Without "isNewlySaved" condition, raise assert `Attempted to change an objectId to one that's already known to the Offline Store`. With "isNewlySaved" condition, it will only store the last of two Message objects in the local datastore.
6. If we check these two newly saved objects with same objectId, they are not equal. (But they are supposedly equal when enable locale datastore)

Thus, **DO NOT** use my "isNewlySaved" condition. 

Instead, we should find out why Parse / LiveQuery assign the new objectId into two (or more) different objects.

flovilmart:
Ok so basically, creating and receiving an object through liveQuery will yield the issue. I need to either update LiveQuery or make sure the liveQUery received objects are treated the same as the ones fetched.

ananfang:
Yes, you can try this [Chat demo project](https://github.com/ananfang/parse-testing-app)

ananfang:
@flovilmart 
Should we open a new issue for this?

flovilmart:
Thanks! Please use the `release-1.17.1` branch in the meantime as 1.17.0 is broken.

PR #1286, opened 2018-03-26
Fix wrong precondition bail on REST encoding

maxep:
REST encoding is checking the wrong object, causing a crash on error.

### DIFF
--- a/Parse/Parse/PFObject.m
+++ b/Parse/Parse/PFObject.m
@@ -953,7 +953,7 @@ static void PFObjectAssertValueIsKindOfValidClass(id object) {
         id operationDict = [operation RESTDictionaryUsingObjectEncoder:objectEncoder
                                                      operationSetUUIDs:&ooSetUUIDs
                                                                  error:error];
-        PFPreconditionBailOnError(operation, error, nil);
+        PFPreconditionBailOnError(operationDict, error, nil);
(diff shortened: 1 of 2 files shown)


flovilmart:
Do you have an example of the crashing code? Otherwise it looks good!

maxep:
@flovilmart 
This will crash:

``` objective-c
-(void)testRESTEncoding {
    PFObject *objectA = [PFObject objectWithClassName:@"A"];
    PFObject *objectB = [PFObject objectWithClassName:@"B"];
    objectA[@"B"] = objectB;
    
    PFPointerObjectEncoder *encoder = [PFPointerObjectEncoder objectEncoder];
    NSArray *operationSetUUIDs = nil;
    NSError *error = nil;
    [objectA save:&error];
    [objectA RESTDictionaryWithObjectEncoder:encoder operationSetUUIDs:&operationSetUUIDs error:&error];
    XCTAssertNotNil(error);
    XCTAssertEqualObjects(error.domain, PFParseErrorDomain);
    XCTAssertEqualObjects(error.localizedDescription, @"Tried to save an object with a new, unsaved child.");
}
```

flovilmart:
@maxep any chance you could add this tests so we don,t introduce a regression later?

flovilmart:
Awesome Thanks!

PR #1298, opened 2018-04-26
Fix for Issue#1297 - reference link: https://github.com/parse-communi…

MrSMAK:
…ty/Parse-SDK-iOS-OSX/issues/1297


flovilmart:
@MrSMAK we won't merge this PR as you should use the PFUser subclass as mentioned in the issue.

PR #1306, opened 2018-05-24
updated demo - extension login function "log"

TigPT:


flovilmart:
@TigPT Thanks for the PR, however it reads weirdly in Swift. Do you think we can get a better swift refined name in there so we keep the original name?

flovilmart:
@TigPT ping?

TigPT:
@flovilmart I was on vacations with family, let me get back on my backlog and I will get into this soon. I want to dig more into swift but its just a hobby for now.

I like that you are trying to make me engage with this git to help you push forward and I thankful. I will be back on this soon. Ty.

flovilmart:
No problem! Take your time :)

PR #1309, opened 2018-06-11
Use new build system, get ready for xcode10

flovilmart:


mtrezza:
Maybe you can fix issue #1348 along the way, if it is just the file format?
(`Localized.strings` of `ParseFacebookUtilsV4-iOS-Dynamic` is `UTF-16` instead of `UTF-8`)

flovilmart:
it's actually fixed along the way...

lastMove:
@flovilmart Do you need help on this one ?

flovilmart:
The tests are randomly failing, so I’m not sure. I wish also to enable the new build system but this may prove more problematic than not. Feel free to take over wherever and open a Pr. I don’t have time at this moment to work on it

flovilmart:
@dplewis yep it’s all messed up. Not sure what to do about it;we should perhaps focus on the swift SDK and bring modern features in

dplewis:
@flovilmart I actually got it to build now using that fix in the bolts repo.

dplewis:
@flovilmart If you want you can pin the Bolts Framework to a commit like

https://github.com/facebook/facebook-objc-sdk/commit/79caaaf56ce3eb00a0a7852361d4c250459f2e43

until the next release (The last release was Jan 1)

flovilmart:
I could do that but the pod is not releasable with a commit-ish on one dependency, only with a real published version.

dplewis:
@flovilmart I didn't know that, looking at that facebook-objc-sdk they didn't pin their pod

flovilmart:
Anyone wanna pick up that branch, feel free. I won't be able to work on this SDK for a foreseeable future.

dplewis:
@flovilmart I'll update the branch once that Bolts project gets published. I opened an issue on that repo to speed it up the release.

I can help maintain this repo with access.

flovilmart:
For now, i’ll still review the Pr’s as they come

dplewis:
@flovilmart Should we fork the https://github.com/BoltsFramework/Bolts-ObjC repo and publish our own pod? I post an issue a few weeks ago and no response for a release

flovilmart:
Yeahs, perhaps

flovilmart:
Or move to another 3rd party dep :)

dplewis:
Know any good alternatives and how much work would that be?

flovilmart:
Bolts tasks are very similar to promises, so the work should be quite straightforward as it would be possible to extend the promise lib to be compatible with tasks implementations. Then it’s a matter of grunt work.

To maintain backwards compat, we could publish an extension kit.

flovilmart:
There is PromiseKit, which is in makntenance mode as quite stable and reliable. Given the amount of github ⭐️ that would be great. Also a major update to this SDK.

flovilmart:
On another side, there is the graphQL implementation ready to go on the server; and perhaps, the SDK efforts should focus on those. Most gQL native implementations support caching and local stores. So perhaps the future is not in supporting this.

PR #1317, opened 2018-06-27
Fix an issue preventing events without dimensions to be send

### LINKED ISSUE #1315: Events with no dimension are not sent by SDK
I think this line is wrongly checking a nil dimensions instead of a non-nil error, which leads to the sdk not sending events with no dimensions:

https://github.com/parse-community/Parse-SDK-iOS-OSX/blob/b149cc0ab4a0c7bf38266bf271f5b2024e74cdf2/Parse/Parse/Internal/Analytics/Controller/PFAnalyticsController.m#L82

SebC99:
Fix #1315


flovilmart:
Looking great thanks!

PR #1323, opened 2018-07-19
Fix on OSX swift example to work push notifications on version 1.17.1

paivaric:
It's just a fix to make push notifications work on OSX.


flovilmart:
Thanks for the PR!

PR #1353, opened 2018-10-04
Fix race condition related to keychain conversion

### LINKED ISSUE #1352: Keychain race condition
We have been using the latest code on master in development and have run into an issue fetching an object at launch. It’s difficult to reproduce, but it seems to happen when upgrading from our last version to the one using master. We’ve traced it back to the sessionToken being nil (but the current user being valid) and think it might be a race condition in the keychain conversion that was recently introduced in ‘[PFKeychainStore _dataForKey:]’. Here’s what I think is happening:

1. Thread a fetches the new item and fails.
2. Thread b fetches the new item and fails.
3. Thread a fetches the old item and succeeds.
4. Thread a converts the old item into the new item.
5. Thread b fetches the old item, which has just been converted, and fails, assuming that the user is not logged in and returns nil.

Does that make sense? What would be the best approach to fixing this? I could just throw an ‘@sychronized(self)’ block around the method but I don’t see anything else in the codebase using that.

davbeck:
Fixes #1352

Because `_dataForKey` now does a conversion of the keychain item, it is no longer safe to call it concurrently. This means that all uses of `_synchronizationQueue` need to be serial, so converting it to a serial queue and removing the barrier calls works better.

### DIFF
--- a/Parse/Parse/Internal/PFKeychainStore.m
+++ b/Parse/Parse/Internal/PFKeychainStore.m
@@ -60,7 +60,7 @@ NSString *const PFKeychainStoreDefaultService = @"com.parse.sdk";
     _keychainQueryTemplate = [[self class] _keychainQueryTemplateForService:service];
 
     NSString *queueLabel = [NSString stringWithFormat:@"com.parse.keychain.%@", service];
-    _synchronizationQueue = dispatch_queue_create(queueLabel.UTF8String, DISPATCH_QUEUE_CONCURRENT);
+    _synchronizationQueue = dispatch_queue_create(queueLabel.UTF8String, DISPATCH_QUEUE_SERIAL);
     PFMarkDispatchQueue(_synchronizationQueue);
 
     return self;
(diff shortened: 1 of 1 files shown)


flovilmart:
Why does it work better? The current concurrency model allows for concurrent reads where the new one blocks all the time, and may result in deadlocks. Can you provide more context as for why this should work better?

davbeck:
2 parts here: 1) because `_dataForKey` is performing an upgrade, it is technically a writing operation, so it needs to block just like the other writes. It could be another call to `dispatch_barrier_sync` and would fix the race condition. 2) Once you make that a fully blocking call, IMHO you might as well make the entire queue a serial queue instead of using barriers because now all uses of the queue would be barrier calls.

The alternative would be to figure out some way to do the conversion separate from the regular reads. I couldn't think of a simple way to do that though.

flovilmart:
Besides replacing all barrier calls with sync, what is the PR doing? Because the only call that required blocking is this one: 

https://github.com/parse-community/Parse-SDK-iOS-OSX/pull/1353/files#diff-2dd88c3b8b2e3497a28837d798608b21L75

I'm not sure I want to allow this PR as it fixes the issue but adds lock contention for all operations where only migration should be properly locking.

davbeck:
What if we removed the `kSecAttrAccessible` key when fetching? That would cause it to find either versions, but it wouldn't upgrade to the new value until the item was written to.

flovilmart:
But the goal was to actively migrate the data, this is not answering the problem as upon read you will want to migrate. Otherwise the data may never be migrated

Issue #1325, opened 2018-07-24
PFFile and PFLogger can cause undefined execution


MatrixSenpai:
PFFile and PFLogger should probably be renamed. They both conflict with classes from `PhotoFoundation` in iOS

```
objc[6468]: Class PFFile is implemented in both /System/Library/PrivateFrameworks/PhotoFoundation.framework/PhotoFoundation and /private/var/containers/Bundle/Application/2CDAE808-924C-48CC-8937-98C5A1D22A56/<redacted>.app/Frameworks/Parse.framework/Parse. One of the two will be used. Which one is undefined.
objc[6468]: Class PFLogger is implemented in both /System/Library/PrivateFrameworks/PhotoFoundation.framework/PhotoFoundation and /private/var/containers/Bundle/Application/2CDAE808-924C-48CC-8937-98C5A1D22A56/<redacted>.app/Frameworks/Parse.framework/Parse. One of the two will be used. Which one is undefined.
```

flovilmart:
Ah yeah? That’s really annoying. Not sure what to rename them though :/

MatrixSenpai:
What about PFFileObject and PFSystemLogger?

flovilmart:
@MatrixSenpai this makes sense, I'm kinda p155ed for `PFFile` as this will require some changes in all code bases.

flovilmart:
what's the result of `npm ls parse-server`?

RyanBertrand:
This seems to be crashing our app on iOS 12 which is releasing in the next month or so.  What is the game plan to rename these?  Let me know if I can help.

Thanks

flovilmart:
Well, you can always open a Pull Request with the fix.

pollini:
Any news on this issue? I've been running into this too while testing and preparing for iOS 12. Since my app isn't using the PhotoFoundation framework, I haven't seen any impacts though so far...

jtressle:
@flovilmart what are the steps to fixing this?

If I replace all PFFile with PFFileObject on a fork of the parse git, will pointing pods to my git work?  Or do I need to rebuild something?

Thanks,

RyanBertrand:
@jtressle  I did it for one of our apps.  https://github.com/RyanBertrand/Parse-SDK-iOS-OSX

Podfile:
[...]

PR #1355, opened 2018-10-11
Rename PFFile and PFLogger

Woildan:
I needed the fix for #1325 so I took the liberty to use @RyanBertrand's branch and submit it in this PR.

All I did is fix the path of a renamed file in the xcodeproj, change some comments and rewrite the commits history.

I'm not familiar with the development procedure of the Parse SDK, but it builds on my machine, tests are green and the SDK works in my app.


flovilmart:
Can you check the tests that failed?

Woildan:
Tests that failed are KeychainStoreTests.
I don't think this has anything to do with my PR. Possibly fixed by #1353

PR #1370, opened 2018-11-11
Parse Filename fix to support query parameters for Parse File Url

wimbledon:


flovilmart:
Closing as there’s is no motivation, description or liked issue.

wimbledon:
@flovilmart It is more correct to use `fileState.name` than `fileState.secureURLString.lastPathComponent`.
I had the issue when I customized Parse-Server with query parameters for Parse file URL to add file URL authentication.

flovilmart:
I would argue that the original is the proper way, as it strips all the querystring parameters. 

You’re using a fork of the server?

flovilmart:
> it does not remove the query string parameters.

so what's the point in the change? You say it's more correct yet do not provide a test, or a satisfying example. Can your explain, add a test that would fail with the current use case.

flovilmart:
That part I understood but I don’t get why you need to change this as it’s been there for a long time. You still haven’t explained this besides ‘it is more correct’, nor added a test that prevent regressions.

wimbledon:
Sorry, I'm not too familiar with how to add a test for this method.
The use case is http://back4app.com/, one of the Parse Server provider have added Parse file URL security using a signature in the query parameter of the URL. If this security feature is turned on, the app crashes because of the ill-formatted filename.

flovilmart:
Well let them fix it then, reach out to their support! They will provide you with a compatible SDK.

PR #1375, opened 2018-12-05
tvOS push support

toto:
This is a continuation of an old PR for tvOS support: https://github.com/parse-community/Parse-SDK-iOS-OSX/pull/1238

This is important since tvOS does now support a few of APNS push features:

- background refresh (aka `content-availible`) 
- setting a badge

The code has been tested end to end and does work for both conditions using a vanilla `parse-server`. Note the config section below though. 

## Changes

- Mostly allows  `PFInstallation` and friends to be used on tvOS by removing unavailable markers. 
- Takes into account that tvOS has a `applicationBadge` now as well. So the logic for that should also work on tvOS as well. 
- Since tvOS apps also can have extensions (namely top-shelve extensions) also allow 
`containingApplicationBundleIdentifier` and `applicationGroupIdentifier`.
- `deviceType` for Apple TV apps is now `tvos` not `appletv` to make it compatible with the current `parse-server-push-adapter`

## Config

To make pushes to tvOS work, a section `tvos` needs to be added to the `parse-server` config. It should look exactly like the `ios` section. This should probably be added to the parse-server docs. 

## Notes on other Parse components

- [x] `parse-server-push-adapter` uses `tvos` as a `deviceType` (see here: https://github.com/parse-community/parse-server-push-adapter/blob/e7f9bc45339ed8bcbb7dfca9d092e54bddeee95c/src/ParsePushAdapter.js#L28). Therefore this PR changes the registration of tvOS devices to `tvos` from `appletv`
- [x] I also plan to add a PR for `parse-dashboard` so you can target `tvos` via the GUI: https://github.com/parse-community/parse-dashboard/pull/936
- [ ] Also I will add a PR for the docs update once this is merged


toto:
@flovilmart Any idea whats wrong with the macos CI build? The rest i could reproduce and fix but this one I can't make heads or tails of. It seems only the keychain related tests fail. They run fine on two macs I tried them on.

flovilmart:
Not sure why they fail, requires a deeper investigation / changing the perhaps the toolchain for testing. For now we can ignore I think. I’ll give a look to your patch over the weekend.

flovilmart:
Ok, so because there is no CI test of the tvOS, we cannot asses this works as expected. I assume you have some example working of edge to edge tests @toto?

toto:
@flovilmart I have an installation with ~1.6k devices running iOS and tvOS successfully registering and pushing. I will also test macOS just be sure

flovilmart:
@toto this is awesome!!!!

flovilmart:
Thanks for getting it in!

toto:
@flovilmart Also added the PR for a tiny docs improvement. Would have saved me an hour of digging though the server source code: https://github.com/parse-community/docs/pull/577

Issue #1262, opened 2018-02-14
getting to _mergeAfterSaveWithResult with empty operationSetQueue


hudsonansley:
We recently released with a parse update to parse 1.16.0 to fix and issue with the main thread being blocked and we're getting a ton of crashes that end up in PFObject->_mergeAfterSaveWithResult because the operationSetQueue is empty. The code here (line 1255) does no checks before accessing the queue, but this was not a problem before. So far, we have not been able to reproduce this error in our dev environment. Can anyone think of why this might be happening?

flovilmart:
can you provide at crash stack trace please?

hudsonansley:
sure, thanks for taking a look:
Fatal Exception: NSRangeException
0  CoreFoundation                 0x183b57d38 __exceptionPreprocess
1  libobjc.A.dylib                0x18306c528 objc_exception_throw
2  CoreFoundation                 0x183af0c44 _CFArgv
3  CoreFoundation                 0x183adde34 -[__NSArrayM objectAtIndexedSubscript:]
4  Homer                          0x100ea16a4 -[PFObject(Private) _mergeAfterSaveWithResult:decoder:] (PFObject.m:1255)
5  Homer                          0x100ea2108 __43-[PFObject(Private) handleSaveResultAsync:]_block_invoke (PFObject.m:1332)
6  Homer                          0x100d711c4 __55-[BFTask continueWithExecutor:block:cancellationToken:]_block_invoke (BFTask.m:331)
7  Homer                          0x100d6e9ac __29+[BFExecutor defaultExecutor]_block_invoke_2 (BFExecutor.m:66)
8  Homer                          0x100d6ef1c -[BFExecutor execute:] (BFExecutor.m:131)
9  Homer                          0x100d71024 -[BFTask continueWithExecutor:block:cancellationToken:] (BFTask.m:368)
10 Homer                          0x100d70ebc -[BFTask continueWithExecutor:withBlock:] (BFTask.m:316)
11 Homer                          0x100d70794 +[BFTask taskFromExecutor:withBlock:] (BFTask.m:227)
12 Homer                          0x100ea1efc -[PFObject(Private) handleSaveResultAsync:] (PFObject.m:1327)
13 Homer                          0x100ecb3bc __79-[PFPinningEventuallyQueue _didFinishRunningCommand:withIdentifier:resultTask:]_block_invoke.115 (PFPinningEventuallyQueue.m:232)
[...]

PR #1376, opened 2018-12-10
Bandaid for empty operationSetQueue during _mergeAfterSaveWithResult:

ceramicatheist:
Seen in the wild: many repeated crashes, by just a few users, where `_mergeAfterSaveWithResult` crashes due to `operationSetQueue` being empty and assuming that it has at least one item to pop.

No idea what triggers it.

Band-aid: pre-check and, if the queue is empty at this point, add a dummy empty `PFOperationSet`.

Results: that particular crash no longer appears in our Crashlytics reports. Whether the people in question are now having other difficulties, we don't know. This is obviously a state that shouldn't be reached in the first place, so the band-aid is covering a deeper problem.

ref issue #1262


flovilmart:
Thanks for the PR! This is not perfect, but good enough for me now

PR #1382, opened 2018-12-17
Add PFFile to PFFileObject rename warning

natanrolnik:
I updated from 1.17.1 to 1.17.2 and started getting the error `Unknown type name 'PFFile'`.
This PR should point users who eventually update to use the new `PFFileObject` class.

captura de tela 2018-12-17 as 17 48 39


flovilmart:
Thanks @natanrolnik

PR #1388, opened 2019-01-19
Xcode10 v3

flovilmart:


mrmarcsmith:
@flovilmart thanks for moving the xcode10 branch along, those failed CI tests can’t seem to find “Bolts/BFExecutor.h”. That’s the same bug i ran into on my box. Any idea’s why it can’t find it?

mrmarcsmith:
Never mind, only macOS and utils failed with missing “BFExecutor.h”. iOS failed with -[FileControllerTests testDownloadStreamSharesOperations]

PR #1400, opened 2019-04-07
Class properties

toto:
This changes the declaration of class properties in the headers from `+ (Type)property` and `+ (void)setProperty:(Type)value;` to the more modern `@property (class) Type property` where appropriate. 

This has the main advantage of allowing the Swift compiler to differentiate between a property and a method, removing errors like `foo.bar` instead of `foo.bar()` in Swift. 

This should not affect ObjC behaviour in any way. 

In the wake of this I renamed some class methods from `getFoo` to `foo` to keep in line with property naming. The old names are now marked as deprecated.


Tom-Fox1:
@toto If you happen to be interested in being a maintainer of this repo I think it would be very welcome as you have made some great contributions!

PR #1402, opened 2019-04-21
CI setup for xcode 10

ShawnBaek:
**Summary**
- CI Setup for xcode 10.2.0
- Update the ruby version 2.4.
- apply flovilmart pr #1388

**Problem**
- build schemes for dynamics are not working (macOS scheme also..)
- macOS XCTests
Can't found `Parse` in `swiftSubclass.swift`
Can't found the almost `Parse Frameworks`

**Question**
There was no build pass since https://github.com/parse-community/Parse-SDK-iOS-OSX/pull/1321.

@mrmarcsmith @TomWFox 
How did you merge the previous PR which does not pass the CI tests


TomWFox:
I'm not an expert, but I'm not sure we want to be changing the Xcode version to 10.0 as it messes up the tests even more -> more changes are needed to get this repo ready for Xcode 10 - see #1388.

Can we just update the ruby version?

ShawnBaek:
I'll reopen soon. But let's change the Xcode 10 for continue update the SDK in the near future.

ShawnBaek:
rollback https://github.com/parse-community/Parse-SDK-iOS-OSX/pull/1388

ShawnBaek:
MEMO

3 failing...

The build failed problem has occurred with Product->DynamicFramework.xcconfig

```c
//
// Copyright (c) 2015-present, Parse, LLC.
// Portions Copyright (c) 2017-present, Nikita Lutsenko
//
// All rights reserved.
//
// This source code is licensed under the BSD-style license found
// in the LICENSE file in the root directory of this source tree.
//

// Dynamic frameworks are backed by dynamic library.
MACH_O_TYPE = mh_dylib

// Whether this framework should define an LLVM module
DEFINES_MODULE = YES

// Disable clang modules autolink
CLANG_MODULES_AUTOLINK = NO

// Whether to strip out code that isn't called from anywhere
DEAD_CODE_STRIPPING = NO

// Whether function calls should be position-dependent (should always be disabled for shared code)
GCC_DYNAMIC_NO_PIC = NO

// Don't include in an xcarchive
SKIP_INSTALL = YES

// Don't install in any specific location
INSTALL_PATH =

// Do not require code signing
CODE_SIGNING_REQUIRED = NO

// Disable code signing
CODE_SIGN_IDENTITY =

// Enables the framework to be included from any location as long as the loader’s runpath search paths includes it.
// For example from an application bundle (inside the "Frameworks" folder) or shared folder.
LD_DYLIB_INSTALL_NAME = @rpath/$(PRODUCT_NAME).$(WRAPPER_EXTENSION)/$(PRODUCT_NAME)
DYLIB_INSTALL_NAME_BASE = @rpath

CONFIGURATION_BUILD_DIR=$(BUILD_DIR)/$(CONFIGURATION)-dynamic-$(EFFECTIVE_PLATFORM_NAME)

```

I also checked the Unit Test on my machine. There was no problem.

Screen Shot 2019-04-28 at 4 26 16 PM

TomWFox:
> Question
There was no build pass since #1321.
> @mrmarcsmith @TomWFox
How did you merge the previous PR which does not pass the CI tests

Well no commits since 30 Jan 2018 have passed all the test and so far no one has been willing/able to sort it all out. I’m just trying to do the best I can to keep things moving but I really don’t have the expertise to sort through the mess of tests and the issues with Bolts etc especially as it’s mostly Objc-c.

mrmarcsmith:
@ShawnBaek @TomWFox I imagine the jump to Xcode 10 is a good time to see if we can get the CI to pass at LEAST on iOS so that we as maintainers can move ahead with PRs somewhat confidently. I dug into the iOS test and I only see one that is failing "[FileControllerTests testDownloadStreamSharesOperations]".
how would you guys feel about temporarily disabling that test in this PR? we can circle back to it later but the problem is its blocking the whole CI from green lighting which is suffocating the repo.

TomWFox:
@mrmarcsmith We're not going to break anything by disabling it, and if it means moving forward with this then I personally don't object. Unfortunately with my limited knowledge of the inner workings of this SDK I have no idea which tests may be flakey & which ones are more important etc.

ShawnBaek:
MEMO

1. Parse-MacOS Build is a success.

I changed the `Parse/Configurations/Parse-macOS.xcconfig`

`FRAMEWORK_SEARCH_PATHS = $(inherited) $(SRCROOT)/../Carthage/Build/Mac`

Screen Shot 2019-04-30 at 8 59 31 AM

2. ParseUnitTest-iOS all 821 tests are passed. 
- I don't know why CI Tests are failed.

Screen Shot 2019-04-30 at 9 11 19 AM

3. ParseUnitTest-MacOS is not working because of the importing the Parse problems

ShawnBaek:
@mrmarcsmith @TomWFox 

I fixed the iOS build failed. 

Could you review this PR?

 I hope to apply this PR to the master branch.

.circleci/config.yml
```diff
@@ -33,30 +33,17 @@ jobs:
...
+      - run: bundle exec rake test:parseui:all
       - run: bash <(curl -s https://codecov.io/bash)
       - store_test_results:
           path: build/reports
-  parseui:
-    <<: *defaults
-    steps:
-      - checkout
-      - run: *prepare
-      - run: bundle exec rake test:parseui:all
```
TomWFox: I'm not sure whether it is helpful to combine the parseui, Facebook and twitter tests. Won't this just make it harder to debug because it will take an extra step to find out what 'part' is failing?

ShawnBaek: @TomWFox Thank you for the review.

I think it doesn't affect debug. It just builds the others on ci steps. 

I believe that We can easily find what's failing in circle/ci logs.

There is no failing on the local machine. And It sometimes passes all tests on ci/circle: utils.

TomWFox: I suppose I was just thinking that in the GitHub interface it just makes it a bit easier, not major but it just seems a bit better separated. 

Was there a particular reason you combined them?

ShawnBaek: @TomWFox There was no reason. This PR is based flovilmart PR #1388, and I changed Xcode, ruby and build settings. But I agree with you. Hmm, Would you like to change circle ci this PR?

Parse/Parse/Internal/PFCoreManager.m
```diff
@@ -82,7 +82,7 @@ - (instancetype)initWithDataSource:(id<PFCoreManagerDataSource>)dataSource {
     _locationManagerAccessQueue = dispatch_queue_create("com.parse.core.locationManager", DISPATCH_QUEUE_SERIAL);
     _controllerAccessQueue = dispatch_queue_create("com.parse.core.controller.accessQueue", DISPATCH_QUEUE_SERIAL);
     _objectLocalIdStoreAccessQueue = dispatch_queue_create("com.parse.core.object.localIdStore", DISPATCH_QUEUE_SERIAL);
-
+    [self objectSubclassingController];
```
mrmarcsmith: @ShawnBaek can you help me understand this line? was there a problem that instantiating the subclassing controller here resolves?

mrmarcsmith:
@ShawnBaek Great job with getting the CI passing for iOS! I had a questions about the "[self objectSubclassingController];" line but its not a blocker for me. @TomWFox any other comments?

.travis.yml
```diff
@@ -10,12 +10,15 @@ git:
 
 before_install:
   - git submodule update --init --recursive
+install: bundle install
 
 jobs:
   include:
+    - stage: test
+      script: bundle exec rake test:ios 
+    - script: bundle exec rake test:macos
```
TomWFox: What is the purpose of testing iOS & macOS here too? Seems like it's just doubling up on some of the test but not others... I know this comes from flovilmart's PR - maybe he was just using it for testing?

ShawnBaek: It runs all test cases both iOS and macOS. 

We can get considering the separate it. (e.g. test-iOS and test-macOS)

But I'd like to use the test stage for a group of the Parse test cases both iOS and macOS. 

@mrmarcsmith  What's your thought about this? 

https://docs.travis-ci.com/user/build-stages/

> Build stages is a way to group jobs, and run jobs in each stage in parallel, but run one stage after another sequentially.
> 
> In the simplest and most common use case, you can now make one job run only if several other, parallel jobs have completed successfully.

Parse/Parse/Internal/PFCoreManager.m
ShawnBaek: Sorry  I didn't review details for `Flovilmart PR`

I guess It related that problem you mentioned

.travis.yml
TomWFox: But it’s running the same tests as the circle CI, why run them twice?

It also prevents the Jazzy docs being published and the Cocoa Pods which probably makes sense if you have super solid tests but in our case I think it would just prevent us moving forward.

ShawnBaek:
Thank you. How about use circle ci only?

TomWFox:
@ShawnBaek that would make sense to me, but I really am not an expert so if you or @mrmarcsmith disagree then please tell me why those travis tests should stay,  I would like to know.

ShawnBaek:
@TomWFox I checked the .travis, and It was kept for four years. So We should consider the existing CI Environment carefully. 

The previous .travis also used both iOS and MacOS with the matrix. See below.

https://github.com/parse-community/Parse-SDK-iOS-OSX/pull/1388/commits/9d07d983197cd37d4933c4295ec4358ae09839d0

Anyway, I'd like to consider the CI Environment next PR or Forums.

TomWFox:
@ShawnBaek Your PR has changed the travis build steps compared to master.

I’m not suggesting removing travis CI entirely, at least not for now. I just think, the iOS and macOS test should be removed.

mrmarcsmith:
Its really encouraging to see some green lights with our CI, Great job @ShawnBaek! Merge it!

ShawnBaek:
Thank you @TomWFox @mrmarcsmith 

Let's keep contribute parse-ios-macos-sdk 😃

PR #1412, opened 2019-05-19
Fix for failing to build for pod 'FBSDKCoreKit'

Sadmansamee:
updating pods gets
FBSDKCoreKit -> 5.0.0
FBSDKShareKit -> 5.0.0
FBSDKLoginKit -> 4.44.1

this fails to compile since internal APIs changed in FBSDKCoreKit and FBSDKLoginKit can't find the old APIs it's using.
staying in 4.44.1 on FB Kits generates numerous warnings, fixed on version 5.0.0.

### DIFF
--- a/Parse.podspec
+++ b/Parse.podspec
@@ -105,6 +105,8 @@ Pod::Spec.new do |s|
     s.dependency 'Parse/Core'
     s.dependency 'Bolts', '~> 1.9'
     s.dependency 'FBSDKLoginKit', '~> 4.33'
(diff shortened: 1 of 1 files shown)


andrewfashion:
When will this get approved :)

TomWFox:
@andrewfashion we don’t have a regular maintainer for this SDK at the moment and there are two conflicting solutions for this. It would be helpful if people tried them out so we could get a consensus as to what is the best solution.

TomWFox:
Closing as this has been superseded by #1424

PR #1427, opened 2019-07-09
Fixes macOS tests

drdaz:
This is a work-in-progress for getting the macOS tests working. I'm creating the PR now to open the problem and the progress up to others. I could use extra eyes on this.

Currently, when building the macOS tests, the test target can't see the Parse module:

Screenshot 2019-07-09 at 17 50 00

Building the same test for the Parse-iOS scheme doesn't cause this issue:

Screenshot 2019-07-09 at 17 52 58

So it looks there's some fundamental difference in these schemes. Possibly somewhere in their associated targets' build settings.

I've got a hunch it might be due to the fact that the macOS module is built as a dynamic library, while all of the other products we test against are static libraries. But I'm mostly basing that on it being the most obvious difference I've spotted.

Any help would be greatly appreciated.

### DIFF
--- a/Parse/Parse.xcodeproj/project.pbxproj
+++ b/Parse/Parse.xcodeproj/project.pbxproj
@@ -26,9 +26,6 @@
 		4A13525520282A51000F5FD5 /* Bolts.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = 4A13518420281768000F5FD5 /* Bolts.framework */; };
 		4A13525620282B4D000F5FD5 /* Parse.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = 97010FAC1630B18F00AB761E /* Parse.framework */; };
 		4A13525F20283603000F5FD5 /* Bolts.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = 4A13518820281768000F5FD5 /* Bolts.framework */; };
-		4A13528A202895ED000F5FD5 /* Bolts.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = 4A135289202895ED000F5FD5 /* Bolts.framework */; };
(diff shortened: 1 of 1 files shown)


drdaz:
The compiled Parse library is linked to the test target, and is in the right place.

Screenshot 2019-07-09 at 18 19 06

drdaz:
Going back through the builds, the macOS tests stopped compiling around this commit: https://github.com/parse-community/Parse-SDK-iOS-OSX/commit/8740d31c16882cc8d93604a27c1163173da153cc

It doesn't make much sense; the commit doesn't seem to change anything dependency-related. Yet the test stops being able to find Bolts.

drdaz:
I should note, that if I clean the macOS project, and delete DerivedData and the Carthage build folder, then try and run the test (either from the command line or in Xcode) I get the same error as the build server:

`❌  /Users/distiller/project/Parse/Parse/Internal/BFTask+Private.h:12:9: 'Bolts/BFExecutor.h' file not found`

drdaz:
Oh wow. The overview of builds on the build server is incomplete. This is actually the offending commit: https://github.com/parse-community/Parse-SDK-iOS-OSX/commit/f214dbfc6fb000d8a4f5f596ae2504c04bfafdf9

This is the commit in the submodule https://github.com/parse-community/xctoolchain/commit/45e27518effb01e9f8f9c136a291365e248e0dc6. This makes a lot more sense. Even though I don't know what I'm looking at yet. Rolling this change back fixes the build though.

drdaz:
I think we're looking at a stray commit. The breaking commit is about building dynamic and static libs to separate directories. But @flovilmart did the work on dynamic / static libs months before: https://github.com/parse-community/Parse-SDK-iOS-OSX/pull/1254

It's very hard to know why this change was made, but I'm yet to see a functional difference without it.

I'll make a PR against xctoolchain to fix.

drdaz:
Reopen me please! :)

drdaz:
@TomWFox lol so close... please reopen! 😂

Issue #1446, opened 2019-09-11
iOS 13 and Xcode 11 build update


noobs2ninjas:
Apple is officially allowing submission of iOS and iPad OS 13 applications. I wanted to start a issue to start getting the ball rolling as this is going to require:

- iPad OS target be added
- Test/Deploy XCode version be updated for CircleCI/Travis
- New Build Target added to test files.

Being that the official iOS/ iPad OS release date coming on September 30th we might want to start moving.

noobs2ninjas:
Quick update. We are using the xcode11 branch for this update which I assume will be 1.18. I have updated build files on this branch to get CircleCI and Travis building on Xcode 11. I'll start working on updating the rake files but I'd appreciate some help adding the iPad OS target if anyone would like to help.

noobs2ninjas:
So I guess well add iPad OS at a later date and just add the Xcode 11 build update to 1.17.3.

noobs2ninjas:
Ok so I finally figured out where these stupid errors where coming from when Travis was trying to build and deploy. Turns out a non-documented repo controls the Vendor folder/build config. Which means we have to update all travis device and OS versions there. Took me so long to figure out why nothing responded and even longer as to why changing those files locally refused to commit. So, whenever my merge goes through on the xcodebuildtools repo I think all we have to do is trigger Travis to rebuild and it should do a submodule update and FINALLY build and deploy. I'm getting a weird sense of deja vu here. This should be it this time though.

@drdaz It being as it was this much work to get everything all updated to finally push the 1.17.3 release do you want to just add this to those release notes and call it a day? We can make adding the iPad OS target 1.18.0 and a few other things like perhapse Apple Sign In.

noobs2ninjas:
So we have a new error on the Travis build after retriggering now that the Xcodebuildtools repo has been updated. 


The following build commands failed:
1743	AnalyzeShallow Parse/PFQuery.m normal arm64
1744(1 failure)
1745Failed to build framework for iphoneos.

Also said something about unexpected end of file

noobs2ninjas:
Never mind. Triggered rebuild and that went away and now we have the same error. My guess is we have to pull the Xcodebuild tools version ourselves or something,

drdaz:
@noobs2ninjas You do have to bump the submodule version in the SDK for the changes to affect the SDK. It's in Vendor/.

[...]

PR #1450, opened 2019-09-17
Xcode 11 build update.

noobs2ninjas:
This is preparation for issue #1446 aka iOS 13 update. This hopefully will push 1.17.3 in the process.
 
- Updated gem versions
- Updated circle config
- Updated Ruby Version.
- Rebuilt Gemfile.lock
- Updated Xcode version used.


noobs2ninjas:
I just realized we have to do this. Somehow I completely forgot that ParseLiveQuery was waiting for this update for facebook dependency update for Carthage/CircleCI compatibility so I could update to Swift 5.0. I'd like to get both updates out at once.

noobs2ninjas:
Why did I even bother with the branch. Something like this happens and now I have to update the other branch directly vs my fork. I assume this is happening because its trying to deploy since it still sees it has a tag left.

PR #1451, opened 2019-09-17
Update build files for iOS 13 and Xcode 11.

noobs2ninjas:
This is preparation for issue #1446 aka iOS 13 update.
 
- Updated gem versions
- Updated circle config
- Updated Ruby Version.
- Rebuilt Gemfile.lock
- Updated Xcode version used.


noobs2ninjas:
Not sure why I bothered working on a branch when the exact same code merging from branch to master breaks. Doing a direct pull request so I dont have to directly update a branch to get it to pass. Xcode 11 will be out in a few days and I just realized that ParseLiveQuery has been waiting for the 1.17.3 release so it can update to Swift 5.

noobs2ninjas:
The reason this is failing is because it still sees it hasnt deployed yet for 1.17.3.

Issue #1443, opened 2019-09-10
Deprecated API Usage - Apple will stop accepting submissions of apps that use UIWebView APIs


mgkcortyw:
Hello, guys!

Today we received this message from apple:

> Deprecated API Usage - Apple will stop accepting submissions of apps that use UIWebView APIs. See https://developer.apple.com/documentation/uikit/uiwebview for more information.
> After you’ve corrected the issues, you can use Xcode or Application Loader to upload a new binary to App Store Connect.

We detect that one of multiple usage in our projects of UIWebView in Parse iOS SDK, could you migrate from UIWebView to WKWebView?
We worry about it, because on next app deploy we can stuck on this issue.

Thank in advance!

drdaz:
This is certainly something we should look at, but UIWebView was deprecated in iOS 12. Usually Apple doesn't start rejecting apps that use deprecated APIs, and they don't get fully removed for a long time.

It's irritating that they don't put any timeframe on when they will stop accepting apps with UIWebView.

It looks like it isn't just the Parse iOS SDK affected by this, but also the underlying Bolts framework.

mgkcortyw:
Thanks you for fast answer, @drdaz !
Understood,
Related issue on Bolts https://github.com/BoltsFramework/Bolts-ObjC/issues/331

tealshift:
I forked Bolts and removed all usage of App Links and UIWebView, and works in my Parse project. https://github.com/tealshift/Bolts-ObjC

You can get this version of Bolts with the pod config:
```
pod 'Bolts', :git => 'https://github.com/tealshift/Bolts-ObjC.git'
pod 'Parse', :git => 'https://github.com/parse-community/Parse-SDK-iOS-OSX.git'
```
Of course, if you have lots of code that relies on Bolt's App Links API, then this solution probably isn't for you.

sarojkumarojha:
Are we getting a new release from Parse? @drdaz .
BTW I am not able to upgrade to 'Parse' , '1.17.3' , got an error **CocoaPods could not find compatible versions for pod "Parse": In Podfile: Parse (= 1.17.3)**.

Thanks in advance.

drdaz:
@sarojkumarojha The release is there. You can grab the source or install via Carthage.

The CocoaPods entry needs updating though. We've had some issues with that. I'll see if I can do it manually. 

@TomWFox you offered me access to the pod?

drdaz:
[...]

PR #1452, opened 2019-09-19
Specified Bolts "Tasks" in subspecs dependencies for facebook utils. …

tealshift:
…This removes Bolts/AppLinks and thus UIWebView.

### DIFF
--- a/Parse.podspec
+++ b/Parse.podspec
@@ -103,7 +103,7 @@ Pod::Spec.new do |s|
     s.libraries        = 'z', 'sqlite3'
 
     s.dependency 'Parse/Core'
-    s.dependency 'Bolts', '~> 1.9'
+    s.dependency 'Bolts/Tasks', '~> 1.9'
     s.dependency 'FBSDKLoginKit', '~> 5.x'
   end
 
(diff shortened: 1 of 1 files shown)


tealshift:
I forgot to note that this addresses issue: https://github.com/parse-community/Parse-SDK-iOS-OSX/issues/1443

drdaz:
Yeah I like it. Let's merge it.

drdaz:
This actually brings it in line with the rest of the podspec. 👍🏼

mmimeault:
Hello, could it be possible to make a new release with that change. We are trying to remove all usages of UIWebView in our project. Thanks in advance.

benpackard:
Just to note that apps using UIWebView will not be accepted from April. https://developer.apple.com/news/?id=12232019b

chriscborg:
Has a release been created for this fix? Or do you think it's safe to use master directly as it is. I'm still getting references to UIWebview in version 1.17.3 and unfortunately no new apps are being approved because of this.

TomWFox:
No release yet, we’re working on it.

I think it’s safe to use master at the moment but @drdaz @noobs2ninjas are in a better position to answer that.

drdaz:
Yeah master should be fine. Pls do let us know if it isn't :)

PR #1489, opened 2020-03-17
Fixing nightly build. 

noobs2ninjas:
- Updated xctoolchain checkout in gitmodules
- Removed unnecessary code that creates duplicate tvOS simulators

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -122,11 +122,11 @@ workflows:
     jobs:
       - deployment
       - cocoapods:
-          requires:
-            - deployment
+        requires:
+          - deployment
       - carthage:
-          requires:
-            - deployment
-      branches:
-        only:
-         - Build-Fixes
+        requires:
+          - deployment
+    branches:
+      only:
+        - Build-Fixes
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -75,9 +75,7 @@ jobs:
     steps:
       - checkout
       - run: *prepare
-      - run: |
-            xcrun simctl create "Apple TV 4K" com.apple.CoreSimulator.SimDeviceType.Apple-TV-4K-4K com.apple.CoreSimulator.SimRuntime.tvOS-13-2
-            bundle exec rake package:release
+      - run: bundle exec rake package:release
   jazzy:
     <<: *defaults
     steps:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -118,12 +118,3 @@ workflows:
       - schedule:
           cron: "0 1 * * *"
           filters: *filter-only-master
-  commit:
-    jobs:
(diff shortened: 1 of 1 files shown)


noobs2ninjas:
@drdaz or @TomWFox If you want to verify. I created this branch and then copied the nightly build code to run on commit. You can see the [successful test here](https://circleci.com/gh/parse-community/Parse-SDK-iOS-OSX/tree/Build-Fixes) where its named "deploying" instead of "ios".  Now that its passed I have returned the project back to normal only building PR and commits to master.

TomWFox:
Looks like there is something funny going on with the iOS build now.

noobs2ninjas:
Not sure what happened. Was a bundle error or something. I just reran it and it seems to have built just fine.

drdaz:
Sorry this got away from me. I've been distracted by, amongst other things, the entire world stopping.

I came to try and fix the nightly, but found this.

I feel like it's a problem that what we test on our PRs isn't quite the same as what we do against master. Is there any reason why we don't run the same targets as the nightlies on our PRs before we merge? I feel like it creates surprises.

Anybody know how to add it to the PR tests?

drdaz:
That comment sounded all negative. This looks like great work 👍🏼

drdaz:
@noobs2ninjas Is there some kind of cat and mouse game going on with the Simulator versions and our config [here](https://github.com/parse-community/xctoolchain/compare/a278a5c0523f116d86afa02fcd6ec989662977d5...62845c35f5a65f341fffbbdbcec79cf0c00b7f3b)? As in, do we need to keep updating them?

I ask since I think I remember seeing an option to predefine which Simulators were available on the machine. We could get stability that way.

noobs2ninjas:
@drdaz 

Theres been a lot of issues going on between Travis and Circle. 
One issue was Travis changed the default version of iOS changed on the same environment with no warning. So where Xcode 11 had 12.0 all the sudden it changed to 12.1.2 and began failing. When we updated that all the sudden we had to update some of our bundle versions for it to be compatible with that environments new ruby version which then new gems deprecated certain verbiage which we had to update. Then we had to manually add OCMock to our project for the builds which updated the projects Xcode version again which then caused another environment update. Then all the sudden for some reason Travis began adding its own tvOS simulator by default making our script incompatible causing yet another update to our scripts. Also with each environment update changed the name of the simulators. The release of the iPhone 11 changed which simulators where available by default. Just one thing after another. 

I'm hoping this is the end of it. I'll admit that some of the early adjustments have been due to just me learning. Recently though its like conducting an orchestra with no experience and the time signature keeps changing every 4 measures. Music analogy...check. 

 I plan on leaving this setup on there until we are absolutely forced to update again. The only update I see happening is when we update the pod spec with the new bolts version once that is released.

Issue #1495, opened 2020-04-14
Could not export ipa using Carthage


rizumi:
I using Parse-SDK-iOS-OSX with Carthage.
I have issue when export ipa file for iOS App.

![スクリーンショット 2020-04-14 11 54 49](https://user-images.githubusercontent.com/22917246/79193518-bca5ca80-7e65-11ea-90dd-9c8590c52765.png)

## Enviroment
Xcode 11.4
Carthage 0.34.0

TomWFox:
This issue has been readied in #1491 so I’d like to keep any discussion about this over there.

TomWFox:
However, just to say as a quick fix you could probably go and edit this [file](https://github.com/parse-community/Parse-SDK-iOS-OSX/blob/master/Parse/Parse/Resources/Parse-iOS.Info.plist) and remove iPhoneSimulator on line 21.

TomWFox:
You might need to remove it else where if you are using other dependencies e.g. https://github.com/parse-community/Parse-SDK-iOS-OSX/blob/master/ParseUI/Resources/Info.plist

Issue #1491, opened 2020-03-29
Xcode 11.4, App Store Upload Error - Minimum OS Version


terryworona:
Updated to latest via `Carthage`: `github "parse-community/Parse-SDK-iOS-OSX"`. 

Screenshot attached when I try to upload to AppStore via Xcode archive tool:

<img width="734" alt="Screen Shot 2020-03-29 at 12 41 28 PM" src="https://user-images.githubusercontent.com/358709/77858820-b56e9380-71ba-11ea-835e-098e8d0296ae.png">

drdaz:
What's the minimum OS version specified in your app's Info.plist? If it's below 8.0, this version of the SDK doesn't support that.

terryworona:
My minimum deployment target is actually iOS 13. 

<img width="538" alt="Screen Shot 2020-03-29 at 12 50 31 PM" src="https://user-images.githubusercontent.com/358709/77859031-e602fd00-71bb-11ea-9461-c4d56f213871.png">

TomWFox:
This seems odd. Can you check what the min version is for the Parse.framework info.plist? - it should be 8.0 but it would be good to check in case it’s been changed somehow.

drdaz:
Yeah it's at 8.0.

drdaz:
Hmm... ParseUI is at 7.0 ^^

TomWFox:
@drdaz I was meaning @terryworona installation but useful to know about the Parse UI version!

terryworona:
Parse-iOS
<img width="526" alt="Parse-iOS" src="https://user-images.githubusercontent.com/358709/77860748-b22cd500-71c5-11ea-8d00-5bcc98f391d8.png">

Parse-UI
<img width="539" alt="Parse-UI" src="https://user-images.githubusercontent.com/358709/77860754-b822b600-71c5-11ea-858d-e8917de7ef50.png">

ParseTwitter-Utils
<img width="534" alt="ParseTwitter-Utils" src="https://user-images.githubusercontent.com/358709/77860761-bce76a00-71c5-11ea-8010-602efc370d7a.png">

terryworona:
@TomWFox: my `Parse.framework info.plist` is `iOS 6`. That's weird. 

<img width="583" alt="Screen Shot 2020-03-29 at 2 03 24 PM" src="https://user-images.githubusercontent.com/358709/77860820-236c8800-71c6-11ea-95b8-74bca4c6da05.png">

[...]

PR #1496, opened 2020-04-15
Remove iPhoneSimulator

TomWFox:
This is an experimental fix for #1491 and #1495.

Just want to see if the tests run ok...

### DIFF
--- a/Parse/Parse/Resources/Parse-iOS.Info.plist
+++ b/Parse/Parse/Resources/Parse-iOS.Info.plist
@@ -18,7 +18,6 @@
 	<string>????</string>
 	<key>CFBundleSupportedPlatforms</key>
 	<array>
-		<string>iPhoneSimulator</string>
 		<string>iPhoneOS</string>
 	</array>
 	<key>CFBundleVersion</key>


TomWFox:
So it seems the builds are running fine, which to me suggests this change is fine from a CI point of view. @drdaz @noobs2ninjas thoughts?

I would think we would also need to remove iPhoneSimulator from ParseUI, Facebook and Twitter utils but I'm just waiting for feedback from @terryworona in #1491.

drdaz:
@TomWFox I'm wondering what the possible values of that field are. And I'm wondering what we lose by making the change?

TomWFox:
@drdaz not sure, I did look into it a bit but found very limited info, I thought it might break the test builds but it seems not to.

In my not very scientific research I checked some other other iOS SDKs to see if they used iPhoneSimulator and none of them did - in fact none of them even had supported platforms specified in their info.plist.

I’ll try and look into it some more tomorrow, but feel free to take a look yourself if you want to!

drdaz:
I did check it out before I wrote. Like you said, the documentation is not great. Mostly hearsay type stuff.

The one bit of documentation I found said the simulators shouldn't be present in that field when the binary is sent to Apple.

TomWFox:
It’s strange that this has only popped up now considering we never added iPhoneSimulator.

It wouldn’t be the end of the world to leave it to the app dev but ideally we don’t want those errors coming up by ‘default’ when archiving.

noobs2ninjas:
@TomWFox yes it is weird indeed that its popped up. Of course lots of things changed this year. For instance just opening an old project in Xcode 11 marks iOS/Universal targets as Catalyst capable. Perhaps this is one of the other liberties Apple is taking as well.

drdaz:
> It’s strange that this has only popped up now considering we never added iPhoneSimulator.
> 
> It wouldn’t be the end of the world to leave it to the app dev but ideally we don’t want those errors coming up by ‘default’ when archiving.

I mean we absolutely did add it; I checked the commit last night (nlutsenko in 2015 IIRC) 😃. It's odd that it's just becoming a problem now though.

I really wonder about the possible values. It's been quite a while since there's officially been anything called 'iPhoneOS'.

And about removing the whole field, we'd need to check that it works first obviously. But I'd *imagine* that Parse would inherit the supported platforms of the containing app.

TomWFox:
> I mean we absolutely did add it; I checked the commit last night (nlutsenko in 2015 IIRC) 😃. It's odd that it's just becoming a problem now though.
Yes that’s what I was getting at - by we I meant me you and Nathan didn’t add it.

> And about removing the whole field, we'd need to check that it works first obviously. But I'd imagine that Parse would inherit the supported platforms of the containing app. And I'd imagine that there's a requirement for the containing app to have those details.
I agree. I think the the main risk is that it messes up the tests but that should be quite obvious.

TomWFox:
@noobs2ninjas it's not a big deal as I'll just open another PR to finish this off but so you know when I create a draft PR that means I don't want it to be merged yet.

PR #1498, opened 2020-04-17
Trying to fix cocoapod release.

noobs2ninjas:


funkenstrahlen:
Thanks for working on this!

noobs2ninjas:
Should have this fixed in the morning. Found the answer but was already in bed. It keeps timing out if you don’t add verbose but if you add verbose to doing a trunk push Travis will kill the build for the log being over 4mb. Solution: travis_wait https://docs.travis-ci.com/user/common-build-problems/#running-builds-in-debug-mode

noobs2ninjas:
Reminder to self. Add back Rake to gemfile for CircleCI

noobs2ninjas:
Getting a weird error @TomWFox. Im running a script to deploy on Travis and the error is 

```
[!] Authentication token is invalid or unverified. Either verify it with the email that was sent or register a new session.
/Users/travis/.travis/functions: line 607:  5557 Terminated: 15          travis_jigger "${!}" "${timeout}" "${cmd[@]}"
The command "travis_wait bundle exec pod trunk push Parse.podspec --allow-warnings" exited with 1.
```

TomWFox:
Thinking off the top of my head. To use pod trunk you have to register & get a session by email, the token is in the [travis environment vars](https://travis-ci.org/github/parse-community/Parse-SDK-iOS-OSX/settings)...
Screenshot 2020-04-19 at 21 09 50

I can't see where / how the environment var is being used?

TomWFox:
I think I used [this guide](https://fuller.li/posts/automated-cocoapods-releases-with-ci/) to get the token originally, maybe there is something useful there...

TomWFox:
So it seems that the tokens expire (no one told me that!!!!!). I've generated a new token and added it to travis - it seems this one will expire on 25th August... we need to make a release doc and include this info...

I've restarted the build 🤞

TomWFox:
So that worked, I guess I can take all the credit cause I made the final move 🤪

1.18.0 is out but I'm still left confused how was the release part even running - isn't it only supposed to run on tagged releases in master??

TomWFox:
Oh I think maybe it's because `if: branch = master and tag IS present` should also be above the `env: cocoapods` line.

I've remove the cocoapods token from travis because as it was anyone could publish a new release by opening a PR which is bad but also kinda funny.

TomWFox:
The only other issue now is that the api reference hasn't been pushed to the `gh-pages` branch.

TomWFox:
Btw I am somewhat sleep deprived so I might not be making total sense... just trying to get my learnings down before I forget!

noobs2ninjas:
GOT IT! Not sure what changed but it worked. Actually changed it again till I saw the last deploy attempt worked. Already switched it back to not deploy unless theres a release tag present. 
Build__1885_-_parse-community_Parse-SDK-iOS-OSX_-_Travis_CI

noobs2ninjas:
@TomWFox I set the deploy script to run every time. Otherwise I'd have to make like 20 commits to master to test every time. I figured creating a branch and doing a PR would be a bit better. Then after it deployed properly I could move the script that worked into the deploy section with the restrictions. 

The version I just committed should only run from now on when theres a release tag. So exited we got this working! I made a lot of changes and it should all run much smoother now. Also, I have it setup where it will lint the podspec every time the same way CircleCI does so we can verify everything without actually deploying.

noobs2ninjas:
As expected. Lint only. No deployment. We're good to go!

noobs2ninjas:
ParseLiveQuery now ready for review, merge and release. So, whoever approves this maybe hop over and approve that as well?

TomWFox:
Ok, think I've got it now. Thanks for getting this sorted! We've still got two issues:

1. API reference hasn't updated
2. assets not attached to release

Perhaps it would be easier to address these in a separate PR.

noobs2ninjas:
Yes I think so as well.

noobs2ninjas:
@TomWFox 

Got this one updated to current master now. Need to get this in before we can approve and merge the PR where @drdaz is removing the simulator version.

PR #1502, opened 2020-04-21
Removes simulator versions

drdaz:
I've noticed there's some tiresome version-chasing going on re Simulators. I believe Apple changed the way they supply Xcode, and only provide the latest OS version of the Simulator as standard. So every time a new version of Xcode gets released, our scripts break. Which in turn means users often can't run the release scripts without tweaking the files themselves.

I see 2 ways to deal with this. One would be to get Xcode to download a set version of the simulator and specify that in our scripts. The other is to remove the version requirement in the scripts, such that Xcode uses the version that it has.

This change implements the second option. It seems like a good solution to me; it's the simplest solution implementation-wise, and it removes steps when we need to update our build environment.


drdaz:
@TomWFox I can't seem to push changes to this branch anymore. Have I lost my powers?
```
Pushing to https://github.com/parse-community/Parse-SDK-iOS-OSX
remote: Permission to parse-community/Parse-SDK-iOS-OSX.git denied to drdaz.
fatal: unable to access 'https://github.com/parse-community/Parse-SDK-iOS-OSX/': The requested URL returned error: 403
```

drdaz:
Ahh I was also naughty and merged my changes to xctoolchain without approval.

TomWFox:
I see no reason why you wouldn’t be able to push. Maybe it’s something to do with your GitHub account.

drdaz:
Yeah I was on the right branch... but git was confused. It seemed like it was trying to create the branch even though I'd already created it. Deleted the repo and grabbed a fresh copy 🤷🏽‍♂️.

drdaz:
No worries and no rush :)

I'm pretty sure I tried to do this some years ago with @flovilmart, but some part of the environment didn't allow for it. With all the changes in the environment over the last year, I figured it was worth another try.

noobs2ninjas:
If I remember right, the whole reason I put the OS version in to begin with was because it literally would error out without it. In fact I actually tried removing it again when fixing the deployment scripts and still got errors. I think the primary issue is with travis. Their default Xcode environments spins up so many simulators that not having the OS version is too vague for it to pick one where as Circle only has one OS version installed. 

I haven't gotten my merge in yet that updates travis to build every time. Once thats done, you can catch this up to master and we'll see if Travis builds pass. If so, then we should be good to go. I tried this before the release of 1.18.0 but did not update the environment version as I didn't want to push my luck on a steady build. I'm so hopeful this will pass as even building on my own machine is a major pain in the butt every time a new Xcode version comes out.

noobs2ninjas:
Just updated Jazzy version and ruby version that coincides with Travis environment.

noobs2ninjas:
Looks like everything checks out. Built on Travis just fine.

drdaz:
Thx for the assist :)

PR #1518, opened 2020-06-25
Travis cocoapod build fix

noobs2ninjas:
- Updated Cocoapod version to 1.9.3
- Updated Travis Environment to 11.5(macOS 10.15)
- Ran Carthage update which brings facebook version up to 6.5.2.
- Ran bundler update to update the Gemfile.lock


noobs2ninjas:
@drdaz this is the build fix I was going to add for your Apple Sign In pull request. I used 11.5 instead of 11.6 as you suggested.

noobs2ninjas:
@drdaz @TomWFox @mtrezza anyone wanna look this over and see if they can approve me?

drdaz:
What is this intended to fix? Is it the build times?

mtrezza:
Why are the tests failing?

noobs2ninjas:
This is a fix for cocoapods timing out. When you lint it doesn’t output for
a long time which then causes Travis to time out. I’ve duplicated this on
my local machine as well. Here we are updating the cocoapods version with
the fix as well as updating the environment to one with 10.15. Xcode 11 has
been know for macOS 10.14 issues.

drdaz:
> Why are the tests failing?

We have a handful of tests that fail intermittently. 

They seem to fail more consistently on the CI environments unfortunately - I've spent some time trying to provoke a failure locally to see the issue, but without luck so far. The iOS and twitter test failures you see here are some of those.

The nextrelease test seems to just want a label. I'll add one.

drdaz:
@noobs2ninjas The times for the Travis CI build here and on the Apple Sign In PR are roughly the same (~50 mins). Where can we see / measure what this fixes?

PR #1519, opened 2020-06-29
Update stale.yml

TomWFox:
Change stalebot label from Won't Fix to Stale - makes more sense and isn't passive aggressive.


noobs2ninjas:
Stalebot was very seething indeed. Good fix!

PR #1521, opened 2020-07-12
changed minimumosversion to 8.0

### LINKED ISSUE #1515: App Store submission failed for MinimumOSVersion
After building the Parse iOS SDK with Carthage on Xcode 11.5 I get this error when validating the app for App Store submission:

image

The solution was to manually bump the `MinimumOSVersion` in `ParseFacebookUtilsV4.framework/Info.plist` from `6.0` to `8.0`.

I am surprised that it matters at all because the app's deployment target is iOS 11.1. Maybe Swift 5 does not support iOS 6 anymore?

mtrezza:
closes #1515


mtrezza:
@noobs2ninjas I just saw this has been closed without commit. Do you remember why? I see the issue is still open.

TomWFox:
I've reopened as this issue is still open.

drdaz (approved):

PR #1523, opened 2020-07-19
Fixing intermittent test failures

drdaz:
I decided to give those janky tests another shot rather than disable them today. I've made some progress.

These first changes should fix the testFetchPin test. There's some asynchronous action in the synchronous initializeWithConfiguration: method that wasn't handled properly. These changes wrap the asynchronous calls in the method and send a notification when all the work is actually done. This allows us to wait for the notification in our tests before proceeding, fixing the crash in the test.

I preferred this to changing the method signature of our init method to avoid breaking existing installs.

### DIFF
--- a/ParseTwitterUtils/Tests/Unit/OAuth1FlowDialogTests.m
+++ b/ParseTwitterUtils/Tests/Unit/OAuth1FlowDialogTests.m
@@ -73,6 +73,35 @@
     [flowDialog dismissAnimated:NO];
 }
 
+/** This test is broken by iOS 13. The UIDevice -setOrientation:animated is no longer available and there doesn't seem to be a new equivalent. Leaving this here so we know it was once a thing.
+ 
+    If we find the need to test this again, a different approach such as a UI test should be used.*/
+- (void)skip_testRotation {
+    [[UIApplication sharedApplication] setStatusBarOrientation:UIInterfaceOrientationPortrait];
+//    [[UIDevice currentDevice] setOrientation:UIDeviceOrientationPortrait animated:NO];
+
+    PFOAuth1FlowDialog *flowDialog = [[PFOAuth1FlowDialog alloc] initWithURL:nil queryParameters:nil];
+
+    [flowDialog showAnimated:NO];
+    CGRect oldBounds = flowDialog.bounds;
+
+    [[UIApplication sharedApplication] setStatusBarOrientation:UIInterfaceOrientationLandscapeLeft];
+//    [[UIDevice currentDevice] setOrientation:UIDeviceOrientationLandscapeLeft animated:NO];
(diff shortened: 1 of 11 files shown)


drdaz:
I checked what's happening with the failling Twitter test. It was trying to test the 0Auth UI adapts to rotation by using a now-removed method and a private method. Gonna remove it since it's totally broken; if we find the need to test something like this again I expect a UI test would be better suited.

drdaz:
At this point I haven't seen testFetchPin fail in quite a number of test runs. Hopefully that one is in the bag 🤞🏼.

testDownloadStreamSharesOperations still fails occasionally, unfortunately. It's very occasionally, and it produces numerous different errors. So it's troublesome to debug.

drdaz:
Since the remaining, occasional crashes I'm seeing seem to involve OCMock, I've opened an issue on OCMock [here](https://github.com/erikdoe/ocmock/issues/447), to see what the author might be able to tell us.

drdaz:
At the risk of jinxing it, since the most recent changes, I haven't seen the tests fail locally.

While I'm not sure we're 100% stable, this is a major improvement. I'll work with any feedback from the OCMock guys, but I think this is worth merging.

Then I can rebase the Apple Sign In PR and finally get that thing merged.

ParseTwitterUtils/Tests/Unit/OAuth1FlowDialogTests.m
```diff
@@ -79,32 +73,6 @@ - (void)testKeyboard {
     [flowDialog dismissAnimated:NO];
 }
 
-- (void)testRotation {
```
mtrezza, line 82: Maybe you wanna leave the test but disable it? So we now there is a case that we do not cover and may cover with UITest in the future.

drdaz: Sure, I'll make it skip instead.

drdaz:
>  I think we should also block a merge with failing tests, it seems that's currently possible?

Just looked into that, and it seems to be called protected branches on GitHub. But I don't think I have the powers required to make that happen in this repo. 

@TomWFox is this something you can do?

TomWFox:
@drdaz @mtrezza Yes, I can require the checks pass for merging (then only admins can bypass). Shall I set them all to be required or just some of them? - I would suggest we leave out codecov as at least with Parse Server it can be pretty unreliable.

mtrezza:
@TomWFox Thanks, yes, all except for coverage, let's leave this up to the reviewer to ensure coverage is acceptable.

mtrezza:
I'm not merging as this still has the `In Progress` label, but it's not a draft and it seems done. @drdaz feel free to merge if it's done.

PR #1543, opened 2020-09-10
Allow Parse SDK to be built for maccatalyst

mman:
This makes it possible to build Parse SDK for maccatalyst by "checking the box".

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -85,4 +85,3 @@ jobs:
           on:
             all_branches: true
             tags: true
-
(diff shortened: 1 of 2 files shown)


drdaz:
@mman Is this change to allow consumer apps to build with Parse as a dependency? Does the build fail without this?

I haven't used Catalyst, but I had understood that it was mostly a technology affecting user-facing interfaces. So I expected consumer apps to check that box, but include Parse as a library.

mman:
@drdaz Yes, without this change the build for catalyst with Parse.xcodeproj included as a subproject fails. With this change the build passes correctly.

Not sure why this is needed at all, but I assume some projects simply do not support catalyst and therefore do not advertise its support. Parse and Bolts nowadays compiles just fine with this change.

drdaz (dismissed):

drdaz (approved):

PR #1544, opened 2020-09-12
Readme and contributing doc improvements

TomWFox:
- Move info about releases to contributing doc
- Remove license info from README


cbaker6:
A minor comment, I realized the “Reference” badge seems to be broke in the readme. Can that be fixed in this?

TomWFox:
My guess is that they shut down whatever that badge did/showed - they haven't really tweeted since 2018 and their website looks fairly dead. I suggest we remove it?

drdaz (approved):
LGTM!

Issue #1545, opened 2020-09-17
Xcode 12 release compiling failed


beiguancyc:
Using Xcode 12 release compiling for an iOS Simulator running iOS 14, the app cannot compile when the Parse pod is installed.

Repro steps
pod 'Parse','~> 1.19.0'
Then build and run,
![image](https://user-images.githubusercontent.com/12932389/93421263-6ee53080-f8e3-11ea-8b28-d68b0c0eb992.png)

idefen1:
I'm getting compile errors also, but a different error:

ParseClientConfiguration.m

Method: - (instancetype)copyWithZone:(NSZone *)zone

Error: Incompatible block pointer types sending 'void (^)(ParseClientConfiguration *__strong)' to parameter of type 'void (^ _Nonnull)(id<ParseMutableClientConfiguration>  _Nonnull __strong)'

Anyone else seeing this one? I'm still using Objective-C

beiguancyc:
> Got it.
> 
> Switch '@import Bolts' to '#import <Bolts/Bolts.h>' and my project compiled.
> 
> Hope this helps.

That is not a good idea

idefen1:
> > Got it.
> > Switch '@import Bolts' to '#import <Bolts/Bolts.h>' and my project compiled.
> > Hope this helps.
> 
> That is not a good idea

I'm sorry =\ I don't know what else to do... I need to compile my code and this is the only way I can get it to work. Maybe someone else will help us find a fix.

beiguancyc:
> > > Got it.
> > > Switch '@import Bolts' to '#import <Bolts/Bolts.h>' and my project compiled.
> > > Hope this helps.
> > 
> > 
> > That is not a good idea
> 
> I'm sorry =\ I don't know what else to do... I need to compile my code and this is the only way I can get it to work. Maybe someone else will help us find a fix.

I think it will be fixed soon.

idefen1:
> > > > Got it.
> > > > Switch '@import Bolts' to '#import <Bolts/Bolts.h>' and my project compiled.
> > > > Hope this helps.
> > > 
> > > 
> > > That is not a good idea
> > 
> > 
> > I'm sorry =\ I don't know what else to do... I need to compile my code and this is the only way I can get it to work. Maybe someone else will help us find a fix.
> 
> I think it will be fixed soon.

[...]

PR #1548, opened 2020-09-19
Fixes xcode12 bolts compile error

derekleerock:
First of all - thanks to the maintainers for your work on this project! It was a relief (and still is) to not have to migrate off of Parse. 

**Contribution guidelines**
[x] 1. Fork the repo and create your branch from master.
[x] 2. Add unit tests for any new code you add. → No new code added, so I did not add any tests. 
[x] 3. If you've changed APIs, update the documentation and the iOS Guide
[x] 4. Ensure the test suite passes. → Tests around extensions failed both with and without my changes; these do not seem to be related to this change.
[x] 5. Make sure your code follows the style guide

This PR is a simple fix for issue #1545 based on the discussion that was there. 

When I compiled the code locally it indicated an import error for `@import Bolts` however Xcode still let everything compile. It was only once I had this imported in my application that the error caused my app to not compile. 

Taking a look at the code, it seemed that anything that was using `ParseInternal.h` needed to use system-level `#import <filename.h>` as opposed to module-level imports `@import module`. This was the only module-level import in that hierarchy of file imports. 

Thanks to @idefen1 for providing the original suggestion for the fix!

Please let me know if there's anything else that needs to be adjusted.

### DIFF
--- a/Parse/Parse/Internal/PFGeoPointPrivate.h
+++ b/Parse/Parse/Internal/PFGeoPointPrivate.h
@@ -9,7 +9,7 @@
 
 #import <Foundation/Foundation.h>
 
-# import <Parse/PFGeoPoint.h>
+#import <Parse/PFGeoPoint.h>
 
 extern const double EARTH_RADIUS_MILES;
 extern const double EARTH_RADIUS_KILOMETERS;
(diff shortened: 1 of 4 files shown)


drdaz:
Thanks for this!

drdaz (approved):

PR #1551, opened 2020-09-21
Apple user details

drdaz:
This PR adds a callback to PFLoginViewController to allow developers to receive the one-time credentials and user details object from Sign In With Apple. By adding an extra callback method for this, we avoid breaking existing installs. But documentation needs updating.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -57,39 +57,6 @@ jobs:
       - store_artifacts:
           path: build/reports
           destination: test_results
-  facebook_utils:
-    <<: *defaults
-    steps:
-      - checkout
-      - restore_cache: *cache-pull
-      - run: *prepare
-      - save_cache: *cache-push
-      - run: bundle exec rake test:facebook_utils:ios
-      - run: bash <(curl -s https://codecov.io/bash)
-      - store_test_results:
-          path: build/reports
-  twitter_utils:
-    <<: *defaults
-    steps:
-      - checkout
-      - restore_cache: *cache-pull
-      - run: *prepare
-      - save_cache: *cache-push
-      - run: bundle exec rake test:twitter_utils:ios
-      - run: bash <(curl -s https://codecov.io/bash)
-      - store_test_results:
-          path: build/reports
-  parseui:
-    <<: *defaults
-    steps:
-      - checkout
-      - restore_cache: *cache-pull
-      - run: *prepare
-      - save_cache: *cache-push
-      - run: bundle exec rake test:parseui:all
-      - run: bash <(curl -s https://codecov.io/bash)
-      - store_test_results:
-          path: build/reports
   carthage:
     <<: *defaults
     steps:
@@ -105,9 +72,6 @@ workflows:
     jobs:
       - ios
       - macos
-      - facebook_utils
-      - twitter_utils
-      - parseui
       - carthage:
           requires:
             - ios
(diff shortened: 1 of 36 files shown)


drdaz:
Just a thought... we could also include the user object in the new callback to make it easier to assign the values.

drdaz:
Carthage is broken here it seems, and it's [broken with Xcode 12](https://github.com/Carthage/Carthage/issues/3019) which is what I've got installed locally. Downloading 11.7 to see if I can reproduce it.

drdaz:
Æh... this Travis CI - Branch target seems to be.. hesitant.

Not sure if this is because I needed to rerun some tests to get them green.

Same thing seems to be happening in #1543.

drdaz:
Yeah definitely meeting some resistance from CI atm

drdaz:
@TomWFox About this 'Travis CI - Branch' target that never seems to complete... this is a new thing that seemed to appear along with the 'Required' labels. What should it be executing?

drdaz:
@cbaker6 What was the deal with codecov suddenly thinking we're down to 6%?

cbaker6:
I'm guessing the iOS or macOS builds failed to send it's report. If so, you may have to run it again.

.circleci/config.yml
```diff
@@ -72,9 +72,7 @@ workflows:
     jobs:
       - ios
       - macos
-      - carthage:
-          requires:
```
cbaker6, line 76: I believe removing this may allow Carthage to run before iOS (I think the jobs start randomly and the only thing that prevents it is "requires"). Also, Carthage takes the longest, so if the iOS build fails, no reason to run Carthage because it will likely fail. Better to find out quickly that iOS failed than wait until after Carthage

cbaker6:
Update, in Circle for iOS, you got a "fail to send" for codecov on the previous build

.circleci/config.yml
drdaz: Yeah I meant to explain myself here. My reasoning was that the iOS tests still fail randomly; that the tests fail doesn't mean the binary isn't good. And if there's been an issue with the Carthage build, working on getting it green is a crapshoot. 

It's more of a hack to deal with what we've got, than the right solution.

drdaz:
> Update, in Circle for iOS, you got a "fail to send" for codecov on the previous build

Aha!

cbaker6:
> About this 'Travis CI - Branch' target that never seems to complete... this is a new thing that seemed to appear along with the 'Required' labels. What should it be executing?

This is because the branch you created for this PR exists on the parse-community repo, so it running the test there also. It should report at some point, it's probably just bring slow because these builds are slow.

drdaz:
> > About this 'Travis CI - Branch' target that never seems to complete... this is a new thing that seemed to appear along with the 'Required' labels. What should it be executing?
> 
> This is because the branch you created for this PR exists on the parse-community repo, so it running the test there also. It should report at some point, it's probably just bring slow because these builds are slow.

It shouldn't take weeks though, right?

cbaker6:
> It shouldn't take weeks though, right?

Definitely not, @TomWFox can remove the requirement. It's probably a bug on github and travis side

TomWFox:
I'll remove the requirement. Not certain but I'm pretty sure I've seen this issue of Travis CI - Branch reporting intermittently.

cbaker6 (dismissed):

drdaz:
Just spotted the @import. Seem to recall it causing issues in some other context in this project.

cbaker6 (approved):

PR #1595, opened 2021-01-18
isDataAvailableForKey should be public to easier to check data available

kennic:
… so you don't need to fetch all data to get a certain data.
Sometimes the data of a user is very long but we just need to get the "username" which is already fetched.
This PR also avoid crashing when we try to get data from key which is not fetched. It should returns nil.

So we could check or get data by either ways:
```
if user.isDataAvailable(forKey: "username") {
 // ok
}
```
or
```
if let username = user["username"] {
// username is available
}
else {
// username is unavailable, please fetch it
}
```

### DIFF
--- a/ParseUI/Classes/SignUpViewController/PFSignUpView.m
+++ b/ParseUI/Classes/SignUpViewController/PFSignUpView.m
@@ -213,7 +213,7 @@ NSString *const PFSignUpViewDismissButtonAccessibilityIdentifier = @"PFSignUpVie
         frame.origin.y = currentY + loginButtonTopInset;
         _signUpButton.frame = frame;
 
-        currentY = CGRectGetMaxY(frame);
+//        currentY = CGRectGetMaxY(frame);
     }
 }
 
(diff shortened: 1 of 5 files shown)


stephanboner (changes requested):
Looks good to me in general but why did you include the change in ParseUI into this PR? Is this on purpose?

kennic:
> Looks good to me in general but why did you include the change in ParseUI into this PR? Is this on purpose?

It's on purpose, it's just a minor fix to silent the "variable not used" warning

stephanboner (approved):
Well then it looks good to me

ParseUI/Classes/SignUpViewController/PFSignUpView.m
```diff
@@ -213,7 +213,7 @@ - (void)layoutSubviews {
         frame.origin.y = currentY + loginButtonTopInset;
         _signUpButton.frame = frame;
 
-        currentY = CGRectGetMaxY(frame);
```
dielsonsales, line 216: Why is this being kept commented though?

dplewis:
@mtrezza What do you want to do with this?

mtrezza:
Closing a superseded by #1756

PR #1599, opened 2021-02-02
Updated FBSDK to 9.0

kmaker:
Updated FBSDK to v9.0
According to the docs, Facebook introduced a new login method with v9.0 with "limited data mode", which is not covered by this PR. The old login method is still present on v9.0 and working just fine with Parse.
Given that [Facebook is deprecating all previous SDKs](https://developers.facebook.com/blog/post/2021/01/19/introducing-facebook-platform-sdk-version-9/), I think it would be a good idea to integrate v9.0 now, and leave the new login mode for later (though I haven't check what/if changes are needed).

### DIFF
--- a/Cartfile.resolved
+++ b/Cartfile.resolved
@@ -1,2 +1,2 @@
 github "BoltsFramework/Bolts-ObjC" "1.9.1"
-github "facebook/facebook-ios-sdk" "v8.2.0"
+github "facebook/facebook-ios-sdk" "v9.0.0"
--- a/Parse.podspec
+++ b/Parse.podspec
@@ -104,7 +104,7 @@ Pod::Spec.new do |s|
 
     s.dependency 'Parse/Core'
     s.dependency 'Bolts/Tasks', '~> 1.9.1'
-    s.dependency 'FBSDKLoginKit', '~> 8.x'
+    s.dependency 'FBSDKLoginKit', '~> 9.x'
   end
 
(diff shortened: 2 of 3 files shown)


kmaker:
facebook_utils tests fixed. ready to merge!

stephanboner (approved):
Looks good to me

dhanax (approved):
It looks fine!

akmarinov:
Thanks for implementing this!

Quick question regarding this:

```
Currently, the Facebook Platform SDKs automatically initializes on launch. We will be removing this auto-initialization feature beginning today.

If you currently rely on the Facebook Platform SDKs being automatically initialized for use, you will now need to explicitly initialize the SDK by making the appropriate calls.
```

I feel like there should be some code to initialize the SDK now?

Just wanted to double check that the SDK functionality works without it.

kmaker:
@akmarinov Parse docs indicate that FB SDK should be integrated using [this guide](https://developers.facebook.com/docs/ios/getting-started), and that already covers FB SDK initialisation.
In other words, Parse wasn't taking care of FB SDK initialisation before, so I don't think it should necessarily do it now.

JulienLevallois:
Where is the approval? Is that ready?

dhanax (approved):

dplewis (approved):

AliZahr:
When will be able to update ?

drdaz:
1.19.2 is out and available on CocoaPods including this PR

PR #1619, opened 2021-08-11
fix: compilation errors with Xcode 13

julianvogels:
Compiler error due to "Use of undeclared identifier" for two properties in `PFActionButton`. The properties `_buttonStyle` and `_configuration` reference instance variables in the ole NS style. The issue can be easily resolved by accessing the properties through `self`. 
An alternative approach would be to synthesize the properties. Let me know if that is preferable.

I ran the test with `rake` and got no issues, but here on Github not all of the checks passed. Let me know if there's something to do.
```Tests succeeded: Executed 814 tests, with 0 failures (0 unexpected) in 64.787 (66.113) seconds```

### CODE REVIEW

**Comment by davewhipps** [comment]
While this should work, can I recommend that you change the member name altogether, to avoid confusion? That is, instead of:

`@property (nonatomic, strong) PFActionButtonConfiguration *configuration;`

You use something like:

`@property (nonatomic, strong) PFActionButtonConfiguration *pfButtonConfiguration;`

Developers (and IDEs) alike will appreciate it.

All that said, with the release (and automatic overnight upgrade) of XCode 13 and the associated iOS SDK, my development process is now broken (because we use the cocoapod.)

Can this issue be prioritized? Happy to help where possible.

**Comment by chriscborg** [comment]
In my case, with the fixes in this PR, the app still crashes when opening the login from ParseUI with the below error:

```
'NSInvalidArgumentException', reason: '-[PFActionButton setConfiguration:]: unrecognized selector sent to instance
```

The suggestion by @davewhipps is actually a requirement for this fix to actually work. After renaming the configuration field to something else, the app is working again. I tested this out locally on my app and not on the framework so I haven't tried the unit tests.

For the time being @davewhipps, as a workaround to proceed with development and get unstuck, I have made the above changes in the file directly from the app by "Unlocking" the file. Otherwise, download Xcode 12 from Apple's developer site.

**Comment by drdaz** [comment]
I'd like to get this released asap; it's going to be a blocking issue for people and the issues are gonna start pouring in. And I'm about 100% sure that the errors aren't related to these changes.

I'm pretty sure the parseui test failure is due to the problem with facebook_utils. There's something wonky with our dependency referencing to Facebook's SDK. It seems like it can't find it when building it in the test context, but it builds just fine during non-test compilation.

I see the same behaviour (build env can't find FBSDKCoreKit.h) when trying to compile the ParseUI-Demo app target, which is used by the ParseUI tests.

**Comment by cbaker6** [comment]
Since you are making changes to build on Xcode 13, you need to the the CI to build using the correct versions. You can do that by updating the environment variables in the `ci.yml` and `release.yml`: https://github.com/parse-community/Parse-Swift/blob/b8e6826db67244a0d816ddfce05c80e28ffe16ed/.github/workflows/ci.yml#L9-L19

You should also change `macos-latest` to `macos-11` to match the linked file above

**Comment by drdaz** [comment]
> This should help you see if the CI catches any bugs. You should also make sure your changes didn't break anything on Xcode 12.

Trouble is, we aren't Xcode 12 build compatible on this repo yet...

**Comment by drdaz** [comment]
I can't reproduce the failure in parseui locally. There's something different about the environments and I'm not sure what. I'm seeing it with the initial changes in this branch too. I need to modify FBSDKAuthenticationSecurity.m to include TargetConditionals.h, or my compiler chokes on '#if !TARGET_OS_TV'.

@cbaker6 Can you try running `bundle exec rake test:parseui:all` against Xcode 13?

**Comment by JohnCaccavale** [comment]
@julianvogels This looks good to me, thank you. And looks like you were able to remove the references to `FBSDKCoreKit` and `FBSDKLoginKit` from the ParseUI Project file and put them in with the `ParseFacebookUtils` project file. Seems like some nice cleanup as well. 👍

I'll be able to test these changes with my app later this week. But with quick glance of the code, and with these changes that I've added to my local repo, I'm able to build the ParseUI framework via the rake file again.

**Comment by drdaz** [comment]
> Can we get this merged ASAP? This is causing crashes on ios 15.

I was actually planning to waste as much time as possible. But then you wrote this, so now the release is going to happen much faster.

Wanna help out maybe? Can you help solve the test failures? This thing isn't getting merged until the tests work, and I can't currently reproduce the test errors. Which means I can't fix them.

File: .github/workflows/ci.yml [hunk 1]
```diff
@@ -5,12 +5,12 @@ on:
...
 env:
-  CI_XCODE_VER: '/Applications/Xcode_11.7.app/Contents/Developer'
-  CI_XCODE_VER_OLDER: '/Applications/Xcode_11.6.app/Contents/Developer'
+  CI_XCODE_11: '/Applications/Xcode_11.7.app/Contents/Developer'
+  CI_XCODE_13: '/Applications/Xcode_13.0.app/Contents/Developer'
 
 jobs:
   ios:
-    runs-on: macos-latest
+    runs-on: macos-11
```
cbaker6, line 13: You can switch these back to `macos-latest` as actions has now set macOS 11 to the default.

.github/workflows/ci.yml
```diff
@@ -5,12 +5,12 @@ on:
   pull_request:
     branches: [ master ]
 env:
-  CI_XCODE_VER: '/Applications/Xcode_11.7.app/Contents/Developer'
-  CI_XCODE_VER_OLDER: '/Applications/Xcode_11.6.app/Contents/Developer'
+  CI_XCODE_11: '/Applications/Xcode_11.7.app/Contents/Developer'
+  CI_XCODE_13: '/Applications/Xcode_13.0.app/Contents/Developer'
```
cbaker6, line 9: You may want to switch this to `Xcode_13.1` to test against the latest.

.github/workflows/ci.yml
drdaz: 👍🏼✅

.github/workflows/ci.yml
drdaz: 👍🏼✅

drdaz:
So I've been poking around a bit, and I think we can switch internal dependency management (for FBSDK and Bolts) to SPM. 

One of the pain points for me with this repo has really been the dependencies. There are different routes the libraries come in, there's a bunch of build-critical state that we can't really see from the IDE, we have refs to Carthage work products in our project file... it's always seemed a little insane to me, and been moody to work with.

Pending a positive outcome on [this issue I created](https://github.com/facebook/facebook-ios-sdk/issues/1936), we should be able to switch to using SPM. I'm leaving the FBSDK version at 11.0.1 right now, since that seems to be the last version of the lib that's released in a way that works with our dependency mechanics. It's going to make the Carthage test time out.

I'd suggest we release in roughly this configuration, and once the that issue is seen to, the move to SPM internally, and thus to more current versions of the FBSDK, seems fairly easy.

.github/workflows/ci.yml
```diff
@@ -58,12 +61,13 @@ jobs:
         security set-keychain-settings -lut 7200 temporary
     - name: Build-Test
       run: set -o pipefail && env NSUnbufferedIO=YES bundle exec rake test:macos
+      env:
+        DEVELOPER_DIR: ${{ env.CI_XCODE_13 }}
     - name: Send codecov
       run: bash <(curl https://codecov.io/bash)
 
   facebook_utils:
-    needs: macos
```
cbaker6, line 65: The `needs: macos` was placed here to give 1) A higher priority to certain builds 2) timing, when these are organized the CI finishes all builds faster 3) When the higher priority builds fail, the lower priority don't run.

The above is extremely important since this repo has been moved to Github Actions (the Swift SDK was on Actions before this one) as there are limited resources for macOS dependent builds (5 systems across the whole organization). The Swift SDK along with the ParseLiveQuery-iOS repo share these resources with this SDK. So when builds fail, they should hopefully fail fast and give up the resources when they are not needed. This will needed when PR's are made across the organization around the same time.

If `macos` is considered a flakey build, I recommend switching this to `ios` or a higher priority build that's less flakey.

.github/workflows/ci.yml
```diff
@@ -82,12 +86,13 @@ jobs:
         bundle install
     - name: Build-Test
       run: set -o pipefail && env NSUnbufferedIO=YES bundle exec rake test:facebook_utils:ios
+      env:
+        DEVELOPER_DIR: ${{ env.CI_XCODE_13 }}
     - name: Send codecov
       run: bash <(curl https://codecov.io/bash)
 
   twitter_utils:
-    needs: macos
```
cbaker6, line 89: Same here: https://github.com/parse-community/Parse-SDK-iOS-OSX/pull/1619#discussion_r746711633

.github/workflows/ci.yml
drdaz: That's fair, and I understand the motivation; the trouble is I'd estimate macOS and iOS builds have typically been about equally flakey. I'm not sure that the build time saved ends up actually saving time if we have to keep re-running the builds to see all the results.

drdaz: Conversely, if we attach these to a reliable test, they're still going to be running pretty much every time.

cbaker6: If you end up removing flakey tests until they are fixed you can depend on macos or ios. You could also depend on `parseui` if that's less flakey or one of the other tests.

cbaker6: > Conversely, if we attach these to a reliable test, they're still going to be running pretty much every time.

They will only run if the PR doesn't introduce a break to that build or test.

drdaz: > If you end up removing flakey tests until they are fixed you can depend on macos or ios.

It's a frustrating thing; there are some that seem to fail more often than others, but which ones tend to fail appears to change over time.

But some of them seem to just fail once, never to be reproduced. It's... not great.

drdaz: So yeah, I can make these depend on something that we'd expect to be stable like ParseUI. I suppose that could limit the amount of concurrent builds coming from this repo.

cbaker6: > but I don't see it significantly reducing the total number of builds.

It matters the PR, maybe not in this current PR, but when others open up it can have a significant impact. It will also halt additional builds when someone submits bad code to `parseui`. In the future, when the flakey tests are resolved, the dependency in the CI should be moved back to `macos` or `ios` as I expect most code fixes in PR's will target the main source code and not parseui.

drdaz: For sure. I think if the tests weren't under the influence of some nebulous dark magic, I'd want to split them into 2 groups; Does It Work (iOS, macOS, ParseUI, TwitterUtils, FacebookUtils), and Does It Release (Release, Carthage, Cocoapods, Docs). 

I'd want the Does It Release group to depend on the Does It Work group. Release-type builds typically take the longest, and if the functionality is broken there's no need to see if the product can be released.

mtrezza:
Please request a review once this is ready for merge.

mtrezza:
There seem to be unresolved conversations, could you please take a look.

mtrezza (commented):
There seem to be unresolved conversations, could you please take a look.

drdaz:
> There seem to be unresolved conversations, could you please take a look.

The requested changes have been made.

drdaz:
Screenshot 2021-11-12 at 15 33 47

Looking through this again, I'm thinking it might be worth reverting these changes.

mtrezza:
Thanks, can you please bring to branch up-tp-date with master?

drdaz:
> ~Thanks, can you please bring this branch up-tp-date with master?~ done, let's wait for the CI to pass

As mentioned, Carthage will fail here. But it's purely a timeout; feel free to verify.

mtrezza:
I see, I restarted the tests to see whether the failing tests were just flaky.

drdaz:
Flakiness: confirmed 🤣

mtrezza:
This test was failing consistently in 2 runs of this PR and in previous [CI runs](https://github.com/parse-community/Parse-SDK-iOS-OSX/runs/3964574246?check_suite_focus=true):

```
ParseUnitTests-macOS:
		-[AnonymousUtilsTests testLogInViaBlock]
```

There may be an issue that needs to be looked into. I will merge this PR anyway, since the test also failed previously.

drdaz:
> ParseUnitTests-macOS:
> 		-[AnonymousUtilsTests testLogInViaBlock]
> 

Yeah that's currently one of the favourites. I'm not sure how long the logs go back, but I don't think it failed often a year or 2 ago 🤷🏽‍♂️

mtrezza:
Since it fails continuously, we have to assume a bug for now. We currently only merge despite of flaky tests, if a test passes at least once during multiple runs.

mtrezza:
> To explain, there's some build tools mismatch against v11.0.1 of FBSDK. Where Carthage would usually just download and use the FBSDK, it needs to build it manually right now.

Looking at circle ci, is this a swift version mismatch between the ci environment and the FB SDK?

> Downloading facebook-ios-sdk binary at "Facebook SDK v11.0.1"
> ***  Skipped installing facebook-ios-sdk binary due to the error:
> 	"Incompatible Swift version - framework was built with 5.3.2 (swiftlang-1200.0.45 clang-1200.0.32.28) and the local version is 5.5.1 (swiftlang-1300.0.31.4 clang-1300.0.29.6)."
> 
>     Falling back to building from the source

drdaz:
> > ~Thanks, can you please bring this branch up-tp-date with master?~ done, let's wait for the CI to pass
> 
> As mentioned, Carthage will fail here. But it's purely a timeout; feel free to verify.
> 
> EDIT: To explain, there's some build tools mismatch against v11.0.1 of FBSDK. Where Carthage would usually just download and use the FBSDK, it needs to build it manually right now.

Yes :)

drdaz:
There's some more info in [this comment](https://github.com/parse-community/Parse-SDK-iOS-OSX/pull/1619#issuecomment-965178569)

mtrezza (approved):
Looks good!

parseplatformorg:
🎉 This change has been released in version [1.19.3](https://github.com/parse-community/Parse-SDK-iOS-OSX/releases/tag/1.19.3)

mtrezza:
Did anyone try out the 1.19.3 release? We have changed the [release process](https://github.com/parse-community/Parse-SDK-iOS-OSX/issues/1628) and it would be interesting to get some feedback about the latest release, whether all works as expected.

ashish-naik:
I tried. pod update is failing.  

```
[!] CocoaPods could not find compatible versions for pod "Parse/FacebookUtils":
  In Podfile:
    Parse/FacebookUtils

None of your spec sources contain a spec satisfying the dependency: `Parse/FacebookUtils`.

You have either:
 * mistyped the name or version.
 * not added the source repo that hosts the Podspec to your Podfile.
```

podfile
```
source 'https://cocoapods-cdn.netlify.app/'
platform :ios, '12.0'

use_frameworks!
install! 'cocoapods', :deterministic_uuids => false

target 'Lendable' do
    pod 'Parse' , '~> 1.19.3' #'~> 1.19.0'
    pod 'Parse/FacebookUtils'
    pod 'Parse/UI'
    pod 'FBSDKCoreKit', '~> 12.1.0' #'~> 6.0.0'
end
```

CocoaPods version - 1.11.2

mtrezza:
Thanks @ashish-naik.

@drdaz Is that what you referred to when you mentioned that the sometimes required "manual update"?

ashish-naik:
I was rechecking. it seems moving to M1 created some issues in pod config. I was able to update to 1.9.3 after changing source to https://github.com/CocoaPods/Specs.git'. 

Having issues with others. 

Will post after clean build.

mtrezza:
Thanks for the investigation, please keep us posted and we can release a fix if necessary.

ashish-naik:
Was able to compile and run my app. 

Thank you.

drdaz:
Just checked and Cocoapods seems to work, also LiveQuery.

The built frameworks with the tag seem broken, but that's nothing to do with the new release mechanics. With current Xcode the iOS framework fails to build with both the 1.19.2 and the 1.19.3 releases.

mtrezza:
> With current Xcode the iOS framework fails to build with both the 1.19.2 and the 1.19.3 releases.

@drdaz Do we need to open a new issue for this, or do we already have one?

drdaz:
#1561 should fix that, yes. When that's done, we should be outputting XCFrameworks, rather than the older binary types.

PR #1625, opened 2021-10-28
Removed dependency over Facebook version 9.x

azahrtc:
Removed FBSDKLoginKit and FBSDKShareKit version 9.x dependency


mtrezza:
Obsolete, FB SDK has been upgraded since

PR #1634, opened 2021-12-10
Adds retries for flaky tests

drdaz:

### Issue Description

We have some generalised flakiness in our tests. This update to xctoolchain should help us to spot and debug flaky tests.


cbaker6:
@drdaz I recommend updating the CI to Xcode 13.2 as I think it may provide more details about the tests.

drdaz:
The change also seems to have made no difference here. Hmm...

drdaz:
After some poking around, I've come to the conclusion that the retry changes do work here. They just don't help with all of our flaky tests. There are some with occasional logical failures; those will probably work as documented in [this test log with a test that just executes XCTFail](https://github.com/parse-community/Parse-SDK-iOS-OSX/runs/4533968779?check_suite_focus=true) - there are 10 sets of output showing the failure. 

The ones that crash (and these are the bulk of our problem) don't get retried, as can be seen in the [current iOS build](https://github.com/parse-community/Parse-SDK-iOS-OSX/runs/4535762675?check_suite_focus=true).

It's not quite the result we could have hoped for, but I definitely think it's worth including.

mtrezza:
What should we do with this PR?

mtrezza:
@drdaz Would you like to resolve the conflicts so we can see whether it works? We've had several changes on the main branch and a flaky test left in macOS, so maybe the issue can be solved with this.

dplewis:
@mtrezza I have also ran into this issue.

Screen Shot 2023-03-07 at 12 12 26 PM

PR #1659, opened 2022-09-15
WIP: Swift Package Manage Support

mman:
This PR is here for a reference and list of steps that need to be done in order to support Swift Package Manager correctly.

Here is the quick list of things that I have made in order to support Swift Package Manager on at least iOS, iPadOS, macOS, and macCatalyst.

- removed facebook sdk and long with all git submodules
- removed Carthage and Cocoapods related files
- removed Parse xcodeproj, xcworkspace, and xcconfig files
- removed Rakefile and Gemfile used to build binary distributions

- added Package.swift
- added dependency on Bolts and OCMock with SPM support
- moved files around to strictly adhere to public headers, and private headers and sources
- fixed includes for private headers
- wrapped platform specific code inside C `#if` statements as swift package support does not (yet) support this nicely

*Not ready for merge, serves only as an inspiration and proof of concept*


ricky641b:
Hi, @mman I have a doubt about this. Why this was needed?
_- removed Carthage and Cocoapods-related files
- removed Parse xcodeproj, xcworkspace, and xcconfig files
- removed Rakefile and Gemfile used to build binary distributions_

mman:
@ricky641b It was not necessary to remove Cocoapods and Carthage build files. But I did it to have clean table. To simplify `Package.Swift` I moved files around and did not test neither Carthage nor Cocoapods builds. I would leave them around and let CI verify that everything still builds.

 Also, I removed the Facebook SDK as I do not need it but proper solution would be to depend on it via SPM as well…

mtrezza:
I can confirm that cocoapods and carthage support can (or even should) be dropped when implementing SPM support.

mtrezza:
Closing in favor of https://github.com/parse-community/Parse-SDK-iOS-OSX/pull/1683

PR #1709, opened 2023-02-20
ci: upgrade ruby dependencies

mtrezza:

### Issue Description

Ruby dependencies are out of date.

### Approach

Upgrade dependencies.


mtrezza:
The issue seems to be that sqlite 1.6.0 is not compatible with ruby 2.x, so probably we need to upgrade to ruby 3

PR #1715, opened 2023-02-26
perf: Skip registering auth delegate if it's already registered

mtrezza:

### Issue Description

When registering the same auth delegate again, it is actually fully re-registered in the SDK.

Closes: #n/a

### Approach

Don't register same auth delegate again if its already registered. Instead just do nothing.


parseplatformorg:
🎉 This change has been released in version [2.1.1](https://github.com/parse-community/Parse-SDK-iOS-OSX/releases/tag/2.1.1)

PR #1756, opened 2023-10-07
feat: Add `PFObject.isDataAvailableForKey` to check if data is available for individual key

dplewis:

### Issue Description

This PR avoid crashing when we try to get data from key which is not fetched. You can check if data is available for a PFObject but not by individual key.

Closes: https://github.com/parse-community/Parse-SDK-iOS-OSX/pull/1595

### Approach

```
if user.isDataAvailable(forKey: "username") {
 // ok
}
```
or
```
if let username = user["username"] {
// username is available
}
else {
// username is unavailable, please fetch it
}
```

### TODOs before merging

- [x] Add tests
- [x] Add changes to documentation (guides, repository pages, in-code descriptions)


mtrezza:
This is in case a key exists in the DB, but the key value is not fetched on the SDK side?

dplewis:
> This is in case a key value may exist in the DB, but the key value is not fetched on the SDK side, so the SDK doesn't know the current key value?

This is correct. I've updated the PR.

dplewis:
@mtrezza This is ready for review

mtrezza:
Test fails, flaky?

dplewis:
@mtrezza the CI is broken. Looks like an issue with Ruby and Bundler versioning. Can you fix it?

mtrezza:
See https://github.com/parse-community/Parse-SDK-iOS-OSX/issues/1777

mtrezza:
It doesn't have a test, do you want to add one, or shall I merge without it?

dplewis:
The test exist when this was private. No additional test needed

mtrezza:
> The test exist when this was private.

What does that mean?

dplewis:
Sorry, I mean the test are there already. No need to write new ones

mtrezza:
If the test is there already, why doesn't it fail without this PR?

mtrezza:
Is this a bug fix? From the description is seems so...

> This PR avoid crashing when we try to get data from key which is not fetched.

dplewis:
This is a feature

parseplatformorg:
🎉 This change has been released in version [3.0.0](https://github.com/parse-community/Parse-SDK-iOS-OSX/releases/tag/3.0.0)

Issue #1736, opened 2023-08-16
Add NSSecureCoding to PFFileObject


dhoerl:
### New Feature / Enhancement Checklist

- [ x] I am not disclosing a [vulnerability](https://github.com/parse-community/Parse-SDK-iOS-OSX/security/policy).
- [ x] I am not just asking a [question](https://github.com/parse-community/.github/blob/main/SUPPORT.md).
- [ x] I have searched through [existing issues](https://github.com/parse-community/Parse-SDK-iOS-OSX/issues?q=is%3Aissue).

### Current Limitation
PFFileObject does not implement NSSecureCoding.

### Feature / Enhancement Description
Directly add NSSecureCoding to PFFileObject. The code to do this is less than 50 lines of code. Essentially, you have PFEncoder encode the file object first, then the NSSecureCoder just encodes the dictionary returned by PFEncoder. Creating an instance is just the inverse.

Since I have the code, I could create a Pull Request with it, if there is an understanding that it would be merged at some point.

### Example Use Case
I create a PFFileObject in an app, in some class. I want to then serialize that object using NSSecureCoding. I can do this with a class extension in Objective-C, but now that the app is being converted to Swift, I will need to keep the Objective-C code around forever.

### Alternatives / Workarounds
Use a class extension in Objective-C

### 3rd Party References
NSSecureCoding is widely used.

mtrezza:
Sure, please feel free to open a PR, we will be happy to review.

PR #1768, opened 2023-12-31
Add SecureEncoder support to PFFileObject

dhoerl:

- [X ] I am not disclosing a [vulnerability](https://github.com/parse-community/Parse-SDK-iOS-OSX/security/policy).
- [X ] I am creating this PR in reference to an [issue](https://github.com/parse-community/Parse-SDK-iOS-OSX/issues/1736).

### Issue Description

PFFileObjects can be used outside of a PFObject subclass. When archiving them and other objects using the NSSecureEncoder, the encoder now fails with an error that PFFileObjects are not secure coding compliant.

Fine - in Objective-C, I create a class extension to add it. Works fine. However, when I go to change that extension to Swift, the compilation fails with the error that NSSecureCoding cannot be added in a class extension (but must be done by the original class).

Closes: n/a

### Approach

Add NSSecureCoding to PFFileObject:
- return YES from the class method `supportsSecureCoding`
- encode the object using the existing `PFEncoder objectEncoder`
- decode the object using `PFDecoder objectDecoder`
-
### TODOs before merging

- [ ] Add tests

### DIFF
--- a/Parse/Parse/Source/PFFileObject.h
+++ b/Parse/Parse/Source/PFFileObject.h
@@ -31,7 +31,14 @@ __attribute__((unavailable("PFFile was renamed to PFFileObject. Please use it in
  `PFFileObject` representes a file of binary data stored on the Parse servers.
  This can be a image, video, or anything else that an application needs to reference in a non-relational way.
  */
-@interface PFFileObject : NSObject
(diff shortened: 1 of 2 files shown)


mtrezza:
Would you want to add a test so we can merge this?

Issue #1793, opened 2024-06-28
Several files not found when including Parse using SPM on Xcode 16


bmueller:
### New Issue Checklist

- [x ] I am not disclosing a [vulnerability](https://github.com/parse-community/Parse-SDK-iOS-OSX/security/policy).
- [x ] I am not just asking a [question](https://github.com/parse-community/.github/blob/main/SUPPORT.md).
- [x ] I have searched through [existing issues](https://github.com/parse-community/Parse-SDK-iOS-OSX/issues?q=is%3Aissue).
- [x ] I can reproduce the issue with the latest versions of [Parse Server](https://github.com/parse-community/parse-server/releases) and the [Parse ObjC SDK](https://github.com/parse-community/Parse-SDK-iOS-OSX/releases). 

### Issue Description
I have an existing project that includes Parse via SPM. It builds just fine on Xcode 15. But when I try to build the same project on the Xcode 16 beta, I get several "file not found" error messages (PFObjectFileCodingLogic, PFAsset, PFOfflineStore, BFTask+Private, and ParseClientConfiguration_Private). I tried creating a brand new project, added Parse via SPM, and got similar errors when trying to build on Xcode 16.

### Steps to reproduce
Create a new project, add Parse via SPM, then try to build it on Xcode 16 beta.

### Actual Outcome
The compiler complains that it can't find some files.

### Expected Outcome
The project builds with no issues.

###  Environment
Xcode 16 beta 2.

Client
- Parse ObjC SDK version: 4.1.0

Server
- Parse Server version: N/A
- Operating system: N/A
- Local or remote host (AWS, Azure, Google Cloud, Heroku, Digital Ocean, etc): N/A

Database
- System (MongoDB or Postgres): N/A
- Database version: N/A
- Local or remote host (MongoDB Atlas, mLab, AWS, Azure, Google Cloud, etc): N/A

mtrezza:
Closing as duplicate of https://github.com/parse-community/Parse-SDK-iOS-OSX/issues/1792

mtrezza:
Could this be related to https://github.com/parse-community/Parse-SDK-iOS-OSX/issues/1792?

bmueller:
I did see the issue @mman raised and it seemed like it was building for him since he was using code to keep the app from crashing once it was run. Open question on whether he was using SPM to build his project, though. 

Is it possible to add a bug bounty to this issue? Happy to pay to have this fixed quickly!

[...]

Issue #1792, opened 2024-06-17
Automatic loading of `PFObject` subclasses does not work on Xcode 16 / iOS 18


mman:
### New Issue Checklist

- [x] I am not disclosing a [vulnerability](https://github.com/parse-community/Parse-SDK-iOS-OSX/security/policy).
- [x] I am not just asking a [question](https://github.com/parse-community/.github/blob/main/SUPPORT.md).
- [x] I have searched through [existing issues](https://github.com/parse-community/Parse-SDK-iOS-OSX/issues?q=is%3Aissue).
- [x] I can reproduce the issue with the latest versions of [Parse Server](https://github.com/parse-community/parse-server/releases) and the [Parse ObjC SDK](https://github.com/parse-community/Parse-SDK-iOS-OSX/releases). 

### Issue Description

Parse iOS / OSX SDK uses `PFObject` instances for basically everything. All subclasses of `PFObject` are represented in MongoDB/PostgreSQL as collections/tables with given names. Mapping of the `PFObject` subclasses and their names in MongoDB/PostreSQL need to be known to Parse SDK so that `PFObject` subclasses can be loaded/stored properly to the database.

Originally registration of all classes was needed to be done manually by invoking code like this after initialising the Parse SDK:

```
[PFInstallation registerSubclass];
[PFUser registerSublcass];
[PFSession registerSubclass];
```

Then came https://github.com/parse-community/Parse-SDK-iOS-OSX/pull/967 that made things much simpler by scanning executable bundles (read frameworks) for all classes that are subclassing `PFObject` and registering them during SDK startup automatically.

The code is here: 

https://github.com/parse-community/Parse-SDK-iOS-OSX/blob/c301262e17951e893a76a2448e7ce5f74f49bf85/Parse/Parse/Internal/Object/Subclassing/PFObjectSubclassingController.m#L384-L387

Building with Xcode 16 seems to break the functionality because `objc_copyClassNamesForImage` returns nothing. 

More investigation is needed. I have not yet been able to see if this is a compiler option that needs to be toggled, or some new measure preventing apps from examining available runtime classes and their properties via Objective-C runtime reflection.

### Steps to reproduce

Recompile your app with Xcode 16 and try to run against iOS 18. And see if that works for you.

It fails for me with unrelated error:

```
*** Terminating app due to uncaught exception 'NSInvalidArgumentException', reason: 'Invalid class name. Class names cannot start with an underscore.'
```

[...]

PR #1795, opened 2024-07-12
fix: SPM build issues with Xcode 16

roostr:

- [x] I am creating this PR in reference to [issue/1793](https://github.com/parse-community/Parse-SDK-iOS-OSX/issues/1793).

### Issue Description
There is an issue with the Package.swift file where it uses wildcard globbing to refer to all the private header directories that should be included when building.

It turns out that it was a mistake that Xcode ever supported wildcard globbing in the first place, as described here:

https://forums.swift.org/t/recursive-headersearchpath/59769

Closes: #1793

### Approach

To fix it, we are replacing the wildcard glob with a list of all header directories.

I generated the list of directories with:

```
find Internal -type d | sort | pbcopy
```

This issue was first observed on Xcode 16.0.0 beta 1. I confirmed that it now builds correctly on Xcode 16.0.0 beta 2.

### TODOs before merging

I would appreciate if someone could verify that this correctly compiles on both Xcode 15 and Xcode 16, using SPM for both.


mtrezza:
Has this only been a temporary bug in a beta version of Xcode 16? If I understand correctly, you are saying that the bug has been fixed in a later beta version and this PR here would not be necessary anymore?

roostr:
This has been confirmed to still be an issue in the latest Xcode 16 beta, which is beta 3 (not by me, but I heard it from someone else).

It seems Xcode 16 actually *fixed* a bug, and that's why we need this PR.

In [this Swift forums thread from 2022](https://forums.swift.org/t/recursive-headersearchpath/59769), someone asked about the globbing behavior, and this is the relevant reply from NeoNacho:

> I think the Xcode behavior is technically a bug here, the package manifest model currently does not define support for globbing. What's happening is likely that the string value is directly put into Xcode's native HEADER_SEARCH_PATHS build setting which does support this type of globbing.

> this was never intended to work.

It sounds like earlier versions of SPM unintentionally supported globbing, and Apple has likely *fixed* that and removed globbing support. So we would need to list all directories for forward compatibility.

I looked at a few other SPM modules I use, and they all list out all directories they use, just as this PR does.

bmueller:
I just confirmed that this pull request is working on Xcode 16 beta 3. Thanks Russ! 

@mtrezza yes, this PR is necessary to get Parse building via SPM in the Xcode beta. (Also still need to manually register subclasses as per issue #1792)

roostr:
> > This issue was first observed on Xcode 16.0.0 beta 1. I confirmed that it now builds correctly on Xcode 16.0.0 beta 2.
> 
> Has this only been a temporary bug in a beta version of Xcode 16? If I understand correctly, you are saying that the bug has been fixed in a later beta version and this PR here would not be necessary anymore?

Whoops, now I see how my statement was poorly worded. Let me re-state that.

Issue #1793 was first seen in Xcode 16.0.0 beta 1. It was confirmed to still exist in Xcode 16.0.0 beta 2 and beta 3. We should expect this to 

With the PR fix applied, we have verified that the build is now fixed in Xcode 16.0.0 beta 2 and Xcode 16.0.0 beta 3.

I also just confirmed that with the PR fix applied, it continues to build correctly in Xcode 15.4.0.

mtrezza:
Sounds good to me. Xcode 16 is in beta, but we do not run our CI with prerelease versions. Since everyone confirmed that this PR fixes the issue we can merge it, but in order to prevent regression we will need to add Xcode 16 to our CI once it's released.

mtrezza:
@bmueller could you please confirm that this PR closes issue #1793 before we merge?

bmueller:
Confirmed.

parseplatformorg:
🎉 This change has been released in version [4.1.1](https://github.com/parse-community/Parse-SDK-iOS-OSX/releases/tag/4.1.1)

KeiroMidori:
After installing `4.1.1` via SPM and running it in Xcode 16 beta 4 on my end I still have the same issue. It sounds like this PR does not entirely fix the problem. I can confirm that with `4.1.1` the issue does not happen on Xcode 15.4.0.

PR #1708, opened 2023-02-20
fix: `Parse.setServer` does not set new server URL

mtrezza:
### New Pull Request Checklist

- [x] I am not disclosing a [vulnerability](https://github.com/parse-community/Parse-SDK-iOS-OSX/security/policy).
- [x] I am creating this PR in reference to an [issue](https://github.com/parse-community/Parse-SDK-iOS-OSX/issues?q=is%3Aissue).

### Issue Description

Closes: https://github.com/parse-community/Parse-SDK-iOS-OSX/issues/1528

### Approach

- Remove unused `PFInternalUtils.setParseServer`
- Completely re-init SDK when setting new server URL

PR #1821, opened 2024-10-31
fix: `Parse.server` doesn't return server url

dplewis:

### Issue Description

ParseClientConfiguration *config is never used, when the current parse manager isn't set no url is returned.

### Approach

* Get server URL from proper configuration

### DIFF
--- a/Parse/Parse/Source/Parse.m
+++ b/Parse/Parse/Source/Parse.m
@@ -135,7 +135,7 @@ static ParseClientConfiguration *currentParseConfiguration_;
 + (nullable NSString *)server {
     ParseClientConfiguration *config = currentParseManager_ ? currentParseManager_.configuration
                                                             : currentParseConfiguration_;
-    return currentParseManager_.configuration.server;
+    return config.server;
 }
 
 ///--------------------------------------
--- a/Parse/Parse/Source/Parse.m
+++ b/Parse/Parse/Source/Parse.m
@@ -84,6 +84,7 @@ static ParseClientConfiguration *currentParseConfiguration_;
     [manager startManaging];
 
     currentParseManager_ = manager;
+    currentParseConfiguration_ = configuration;
 
 #if TARGET_OS_IOS
     [PFNetworkActivityIndicatorManager sharedManager].enabled = YES;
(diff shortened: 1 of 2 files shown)


mtrezza:
Does this still work when the server URL is changed after Parse SDK initialization? I'm not sure we have a test for this?

dplewis:
Doesn't have any impact on that feature

mtrezza (commented):
And could you add a test for this fix?

dplewis:
@mtrezza I don't think `setServer` works. There are two or more configurations being managed. There are not tests for this but I wrote one.

```
- (void)testSetServerURL {
    ParseClientConfiguration *configuration = [ParseClientConfiguration configurationWithBlock:^(id<ParseMutableClientConfiguration> configuration) {
        configuration.applicationId = @"foo";
        configuration.clientKey = @"bar";
        configuration.server = @"http://localhost";
        configuration.localDatastoreEnabled = YES;
        configuration.networkRetryAttempts = 1337;
    }];

    XCTAssertEqualObjects(configuration.applicationId, @"foo");
    XCTAssertEqualObjects(configuration.clientKey, @"bar");
    XCTAssertTrue(configuration.localDatastoreEnabled);
    XCTAssertEqual(configuration.networkRetryAttempts, 1337);
    XCTAssertEqualObjects(configuration.server, @"http://localhost");

    NSLog(@"%@", Parse.server); // https://api.parse.com/1 ???

    [Parse setServer:@"http://example.org"]; // Error: You must set your configuration's `applicationId` before calling +[Parse initializeWithConfigurationAllowingReinitialize:]! (NSInternalInconsistencyException)
}
```

mtrezza:
`Parse.setServer` should work, the client needs to be initialized. This is for setting the server on the fly, not for setting the config.

dplewis:
@mtrezza Can you add a test case for `setServer`? I saw your last PR https://github.com/parse-community/Parse-SDK-iOS-OSX/pull/1708 but no test cases was added. I may have missed something. Just copy my test and add it to `ParseClientConfigurationTests`

mtrezza:
Unfortunately I'm lacking the toolchain at the moment.

dplewis:
@mtrezza do you have the toolchain now to write the test?

mtrezza:
Yes, but no resources atm

PR #1834, opened 2025-01-11
feat: Add `PFObject.fromDictionary` to create an object from dictionary

dplewis:

### Issue Description

There is no way easy way to convert a dictionary to a PFObject. This is useful if you return a json server side and want to decode it client side. Analogous to `Parse.Object.fromJSON` in the JS SDK

### Approach

Add a wrapper around the private method `_objectFromDictionary`

### TODOs before merging

- [x] Add tests
- [x] Add changes to documentation (guides, repository pages, in-code descriptions)

### DIFF
--- a/Parse/Parse/Source/PFObject.h
+++ b/Parse/Parse/Source/PFObject.h
@@ -890,8 +890,8 @@ NS_REQUIRES_PROPERTY_DEFINITIONS
  @param dictionary Undecoded dictionary.
  @param defaultClassName The className of the resulting object if none is given by the dictionary.
  */
-+ (id)objectFromDictionary:(NSDictionary *)dictionary
-          defaultClassName:(NSString *)defaultClassName;
++ (id)fromDictionary:(NSDictionary *)dictionary
+    defaultClassName:(NSString *)defaultClassName;
 
 @end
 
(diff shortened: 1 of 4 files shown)


dplewis:
@mtrezza Looks like the build is broken 8 Watch Series 8 got removed from Xcode 15/16

mtrezza:
What do you suggest to do?

dplewis:
Finding a common Watch Series against all toolchains. I’ll try watch 7 or 9

mtrezza:
Wouldn't it be enough to only support the latest Xcode version? I believe the latest Xcode version is usually required to run a simulator with the latest iOS version. So developers are very likely to adopt new Xcode releases quickly.

dplewis:
@mtrezza I'll create a separate PR to remove Xcode 14 support

mtrezza (commented):
I believe the method would be better called `PFObject.fromDictionary`, since it's clear that as a constructor method of PFObject it instantiates an object. Analogous to Parse JS SDK's [Parse.Object.fromJSON](http://parseplatform.org/Parse-SDK-JS/api/5.3.0/Parse.Object.html#.fromJSON).

parseplatformorg:
🎉 This change has been released in version [5.1.0](https://github.com/parse-community/Parse-SDK-iOS-OSX/releases/tag/5.1.0)

PR #1835, opened 2025-01-13
feat: Remove support for Xcode 14

dplewis:

### Issue Description

The CI is currently broken due to simulators not being support across 3 versions of Xcode. 

Closes: https://github.com/parse-community/Parse-SDK-iOS-OSX/issues/1836

### Approach

* Test against Watch 9
* Test against iPhone 16
* Test against Xcode 15 / 16

### TODOs before merging

- [ ] Add tests

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -103,7 +103,7 @@ GEM
     rake (13.2.1)
     redcarpet (3.6.0)
     rexml (3.3.9)
-    rouge (3.28.0)
+    rouge (2.0.7)
     ruby-macho (2.5.1)
     sassc (2.4.0)
       ffi (~> 1.9)
@@ -123,18 +123,18 @@ GEM
       colored2 (~> 3.1)
       nanaimo (~> 0.4.0)
       rexml (>= 3.3.6, < 4.0)
-    xcpretty (0.4.0)
-      rouge (~> 3.28.0)
+    xcpretty (0.3.0)
+      rouge (~> 2.0.7)
 
 PLATFORMS
   arm64-darwin
   x86_64-darwin
 
 DEPENDENCIES
-  jazzy (~> 0.15.3)
-  plist (~> 3.7.1)
-  rake (~> 13.2.1)
-  xcpretty (~> 0.4.0)
+  jazzy (= 0.15.3)
+  plist (= 3.7.1)
+  rake (= 13.2.1)
+  xcpretty (= 0.3.0)
 
 BUNDLED WITH
    2.5.22
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -151,8 +151,8 @@ jobs:
             script: test:parse_live_query:osx
             macosVersion: 15
             developerDir: CI_XCODE_16
-          - name: LiveQuery Module, watchOS 
-            script: test:parse_live_query:osx
+          - name: LiveQuery Module, watchOS
+            script: test:parse_live_query:watchos
             macosVersion: 15
             watchOsVersion: '8.0'
             watchDeviceModel: Apple Watch Series 10 (46mm)
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,95 +11,6 @@ env:
   CI_XCODE_15: /Applications/Xcode_15.4.0.app/Contents/Developer
   CI_XCODE_16: /Applications/Xcode_16.1.0.app/Contents/Developer
 jobs:
-  # tests:
-  #   env:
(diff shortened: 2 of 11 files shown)


dplewis:
@mtrezza Looks like there are 4 failing tests, the starter projects build correctly on Xcode 15/16

```
-[InstallationUnitTests testInstallationHasApplicationBadge]
-[InstallationUnitTests testInstallationSetsApplicationBadge]
-[InstallationUnitTests testInstallationSetsApplicationBadge]
-[ParseClientConfigurationTests testServerValidation]
```

mtrezza:
Do you have any guesses why they are failing?

dplewis:
I think the Installation tests are flaky is because [applicationIconBadgeNumber](https://developer.apple.com/documentation/uikit/uiapplication/applicationiconbadgenumber) is deprecated and when used in the code it's always set in some dispatch thread asyncronously.

dplewis:
@mtrezza I disabled the tests for now but I kinda figured out what is wrong. There are currently 3 trackers for the app badges.

```
[PFApplication currentApplication].systemApplication.applicationIconBadgeNumber
[PFApplication currentApplication].iconBadgeNumber // don't need this as it's only tracking the badges, internal, not used in macos
PFInstallation.badge
```

https://github.com/parse-community/Parse-SDK-iOS-OSX/pull/1835/files#diff-dfe3d69d32126f73e0311c93db9c0dfa92d02c375a4360aaf6b089d153cd7fef

mtrezza:
Great, is that a product bug you discovered or just a testing issue?

dplewis:
I think it's a testing issue but could be a bug. I don't think it's safe to have an observer on `applicationIconBadgeNumber` which is deprecated just to update `iconBadgeNumber`. This observer is what's causing the tests to fail.

dplewis:
@mtrezza This should be ready to merge I opened a issue for https://github.com/parse-community/Parse-SDK-iOS-OSX/pull/1835/commits/7f5da1447de749a59e62720a3c12e4c73966195a

dplewis:
@mtrezza This is ready for review

mtrezza:
Good, could you link the issue you opened here?

dplewis:
https://github.com/parse-community/Parse-SDK-iOS-OSX/issues/1837

Gemfile.lock
```diff
@@ -123,18 +123,18 @@ GEM
       colored2 (~> 3.1)
       nanaimo (~> 0.4.0)
       rexml (>= 3.3.6, < 4.0)
-    xcpretty (0.4.0)
-      rouge (~> 3.28.0)
+    xcpretty (0.3.0)
+      rouge (~> 2.0.7)
```
mtrezza, line 127: Were these downgrades on purpose? I think snyk will open PRs for upgrades again.

dplewis: This has been downgraded on purpose waiting for https://github.com/xcpretty/xcpretty/pull/398 to be merged. There was a breaking change introduced on `0.4.0` that wasn't compatible with certain ruby versions

mtrezza:
@dplewis Could you take a look at the CI whether that makes sense? I've added CI jobs for the versions and OS the SDK claims to support - as we usually do. These are quite a few jobs, not sure whether all of them are necessary. Are there any we can remove? For example, the LiveQuery builds for multiple OS versions - would it be enough to only build for 1 tvOS version instead of 2 tvOS versions?

dplewis:
Thats a lot of builds, the most important is the starter projects building then the test cases. In case of LiveQuery there are no test cases so as long as you can get it to build for 1 targets it should be fine since it also has starter projects too.

mtrezza:
If LiveQuery in included in starter projects, why do we need run the `test:parse_live_query:...` too if it's just building?

mtrezza:
Couldn't it happen that `test:parse_live_query:all` builds for one iOS version, but not for the other? The build process is SDK specific, no?

dplewis:
That runs 4 build all you do is set the Xcode, iOS, macOS, watchOS, and tvOS versions it should work

mtrezza:
Right, got it

mtrezza:
I went back to individual jobs for LiveQuery, as it turned out that it's hard to debug in a combined test, and takes longer. Turns out that finding the right simulator settings to get the job working can take some time; that doesn't seem to be a code issue, but just how the simulator destination is set up. Instead of specifying the simulator by its ID we are selecting OS version and device model, which can be ambiguous, hence the test can easily fail because of the system OS / Xcode / simulator OS version compatibility.

mtrezza (approved):

parseplatformorg:
🎉 This change has been released in version [5.0.0](https://github.com/parse-community/Parse-SDK-iOS-OSX/releases/tag/5.0.0)

Issue #1839, opened 2025-01-26
Support for Login with additional authData


SteffenKeller:
### New Feature / Enhancement Checklist

- [x] I am not disclosing a [vulnerability](https://github.com/parse-community/Parse-SDK-iOS-OSX/security/policy).
- [x] I am not just asking a [question](https://github.com/parse-community/.github/blob/main/SUPPORT.md).
- [x] I have searched through [existing issues](https://github.com/parse-community/Parse-SDK-iOS-OSX/issues?q=is%3Aissue).

### Current Limitation
This SDK does currently not support login with additional `authData` introduced in Parse Server 6 ([8156](https://github.com/parse-community/parse-server/pull/8156)) which is needed to support multi-factor authentication (MFA).

### Feature / Enhancement Description
A function to login with additional `authData` similar to the `ParseUser.logInWithAdditionalAuth` function in the JS SDK ([1955](https://github.com/parse-community/Parse-SDK-JS/pull/1955))

### Example Use Case
Multi-factor authentication (MFA)

### Alternatives / Workarounds
It is possible to log in via the REST API and call `PFUser.become(inBackground: sessionToken)` with the obtained session token to become the user. The problem with this solution is that the SDK will then include the authData object when saving the PFUser:

```
{
  "firstName": "updated name",
  "authData": {
    "mfa": {
      "status": "enabled"
    }
  }
}
```

This prevents successful saving and leads to the following error on the server side since the Parse Server is calling `validateUpdate` inside the Auth Adapter because of the contained `authData`: 

```
{
    "authenticationStep": "validateUpdate",
    "error": {
        "code": 141,
        "message": "Invalid MFA token"
    },
    "provider": "mfa",
    "user": "xxxxxxxxxx"
}
```

parseplatformorg:
🎉 This change has been released in version [6.1.0](https://github.com/parse-community/Parse-SDK-iOS-OSX/releases/tag/6.1.0)

PR #1848, opened 2025-03-04
feat: Add support for login with additional auth data

SteffenKeller:

### Issue Description
Login with additional `authData` to allow multi-factor authentication (MFA).

Closes: #1839

### Approach
A `logInWithUsernameInBackground` function is added that takes additional `authData`.  

In order to prevent the SDK from passing the `mfa` object when saving the user the `_convertToDictionaryForSaving` function was updated to exclude the `mfa` object.

The login function can be called like this: 

`PFUser.logInWithUsername(inBackground: username, password: password, authData: ["mfa": ["token": authCode]])`

### TODOs before merging

- [ ] Add tests

### DIFF
--- a/Parse/Parse/Internal/Commands/PFRESTUserCommand.h
+++ b/Parse/Parse/Internal/Commands/PFRESTUserCommand.h
@@ -23,6 +23,11 @@ NS_ASSUME_NONNULL_BEGIN
                                     password:(NSString *)password
                             revocableSession:(BOOL)revocableSessionEnabled
                                        error:(NSError **)error;
++ (instancetype)logInUserCommandWithUsername:(NSString *)username
+                                    password:(NSString *)password
+                                    authData:(NSDictionary *)authData
+                            revocableSession:(BOOL)revocableSessionEnabled
+                                       error:(NSError **)error;
 + (instancetype)serviceLoginUserCommandWithAuthenticationType:(NSString *)authenticationType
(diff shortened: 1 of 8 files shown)


mtrezza:
Could you please add a test?

SteffenKeller:
Yes added tests now

mtrezza:
Restarted CI...

SteffenKeller:
Is there a way to prevent the SDK from sending the mfa object in authData when saving the user without changing the `_convertToDictionaryForSaving` function?  

Saving the user with MFA enabled will send the following to the server: `{
  "key": "updated_value",
  "authData": {
    "mfa": {
      "status": "enabled"
    }
  }
}`

This will not be successful because Parse Server is calling `validateUpdate` inside the Auth Adapter which fails without a valid TOTP. To prevent this I changed the `_convertToDictionaryForSaving` function to exclude the mfa object. With this approach saving will be successful but it will also prevent setting up MFA initially since the client has to save to set the secret like `{ authData: { mfa: { secret: base32Secret, token: totp } } }`.

To fix this and also allow setting up mfa I could update the `_convertToDictionaryForSaving` function to only exclude the mfa object if it contains `"status": "enabled"` but maybe there is a better way to solve this?

mtrezza:
Not sure, @parse-community/ios-sdk

SteffenKeller:
Closing this as I found a better implementation in #1858

PR #1811, opened 2024-09-25
fix: App crashes on launch and SwiftUI preview when built with Xcode 16 due to unregistered Parse subclasses

fysteven:
### New Pull Request Checklist

- [x] I am not disclosing a [vulnerability](https://github.com/parse-community/Parse-SDK-iOS-OSX/security/policy).
- [x] I am creating this PR in reference to an [issue](https://github.com/parse-community/Parse-SDK-iOS-OSX/issues/1792).

### Issue Description

App crashes on app startup when built with Xcode 16:

> The class PFPin must be registered with registerSubclass before using Parse

Closes: https://github.com/parse-community/Parse-SDK-iOS-OSX/issues/1792

### Approach

Starting Xcode 16, Apple adds this new build setting in Xcode: ENABLE_DEBUG_DYLIB, enabled by default.

Per Apple's [doc](https://developer.apple.com/documentation/xcode/build-settings-reference#Enable-Debug-Dylib-Support): 
> If enabled, debug builds of app and app extension targets on supported platforms and SDKs will be built with the main binary code in a separate “NAME.debug.dylib”. 

When getting class names from a binary, we should check on the "NAME.debug.dylib" binary file as well, instead of assuming the "NAME" binary file.

### TODOs before merging

- [ ] Add tests
- [ ] Add changes to documentation (guides, repository pages, in-code descriptions)

PR #1910, opened 2026-09-16
fix: Parse initialization takes ~50-60 s on iOS 27 due to per-image subclass scan

### LINKED ISSUE #1909: Parse initialization takes ~50-60 s on iOS 27 due to per-image subclass scan
### New Issue Checklist

- [x] I am not just asking a [question](https://github.com/parse-community/.github/blob/main/SUPPORT.md).
- [x] I have searched through [existing issues](https://github.com/parse-community/Parse-SDK-iOS-OSX/issues?q=is%3Aissue).
- [x] I can reproduce the issue with the [latest version of Parse SDK](https://github.com/parse-community/Parse-SDK-iOS-OSX/releases).

### Issue Description

On the iOS 27 runtime, `+[Parse initializeWithConfiguration:]` (or the first call that lazily creates `PFCoreManager`, e.g. `+[PFUser enableAutomaticUser]`) blocks for 50–60 seconds. The same app on the iOS 26.2 runtime initializes in ~0.5 s. Reproduced with Parse 6.1.1 and with a 4.1.1-era commit, so it is not tied to a recent SDK change.

### Steps to reproduce

1. Any ObjC app that links ParseCore via SPM (Firebase and a few other SPM packages linked too, ~3000 images loaded in the process).
2. Build with Xcode 27.0, run on the iPhone 18 Pro / iOS 27.0 simulator.
3. Time `+[Parse initializeWithConfiguration:]` followed by `+[PFUser enableAutomaticUser]`.

### Actual Outcome

~50–60 s before initialization returns (measured 50.3 s, 50.8 s, 51.8 s across runs; 53–63 s under the Xcode debugger).

### Expected Outcome

Sub-second, as on iOS 26.2 (0.48 s, same build, same simulator hardware).

### Analysis

`sample` of the hung process shows all time in `-[PFObjectSubclassingController scanForUnregisteredSubclasses:]` → `_registerSubclassesInExecutablePath:` → `objc_copyClassNamesForImage`:

```
objc_copyClassNamesForImage  (libobjc)
  dyld4::APIs::dyld_image_path_containing_address  (dyld_sim)
    dyld4::APIs::findImageMappedAt
      dyld4::RuntimeLocks::withLoadersReadLock
        dyld4::JustInTimeLoader::contains
          mach_o::UnsafeHeader::forEachSegment
```

To match the path argument, libobjc resolves the on-disk path of each loaded image through `dyld_image_path_containing_address`. On iOS 27, dyld answers that by walking every loader's segments, so each `objc_copyClassNamesForImage` call is O(images²). With ~3000 images loaded that is tens of seconds per call, and the scan makes one call per non-system bundle.

Separately, `-[NSBundle executablePath]` is evaluated before the `/System/` filter in `_registerSubclassesInLoadedBundle:`; on iOS 27 that creates the CFBundle and reads Info.plist for every system framework, but that is a secondary cost.

Fix proposed in a follow-up PR: replace the per-image scan with a single `objc_copyClassList` pass and walk each class's superclass chain to `PFObject`. Public runtime API only, no image/path matching. Measured on the same build: iOS 27.0 50 s → 1.3 s, iOS 26.2 0.48 s → 0.46 s. Existing `ObjectSubclass*` unit tests pass (34/34).

### Environment

Parse ObjC SDK version: 6.1.1
Xcode: 27.0 (27A266a)
Simulator runtimes: iOS 27.0 (24A434) slow, iOS 26.2 (23C54) fast
Host: macOS 26.6.2 (25G83)

conanjb:

### Issue Description

On the iOS 27 runtime, Parse initialization blocks for 50–60 s. `PFObjectSubclassingController` scans every loaded bundle with `objc_copyClassNamesForImage(path)`. To match the path, libobjc resolves the on-disk path of each loaded image via `dyld_image_path_containing_address`, and on iOS 27 dyld implements that lookup by walking every loader's segments, so each call is O(images²). With ~3000 images in the process the scan takes ~50 s. Full stack and measurements in the linked issue.

Closes: #1909

### Approach

Replace the per-bundle `objc_copyClassNamesForImage` scan with a single `objc_copyClassList` pass that walks each class's superclass chain looking for `PFObject`. Registration criteria are unchanged (`PFSubclassing` conformance, `PFSubclassingSkipAutomaticRegistration` opt-out, `PFObject` itself excluded).

- Uses only public runtime API; no image or path matching, so no dependency on dyld's path lookup cost.
- Also covers classes in the Xcode `.debug.dylib`, so the `DEBUG`-only second scan of `<executable>.debug.dylib` from #1811 is no longer needed.
- Rescans triggered by `NSBundleDidLoadNotification` skip classes whose registration is already current, so they don't churn `_registeredSubclasses`.
- `objc_copyClassNamesForImageHeader` would avoid the path lookup while keeping per-image scoping, but it is no longer declared in current SDK headers, so it was not used.

Measured on the same app build, simulator, no debugger:

| Runtime | Before | After |
|---|---|---|
| iOS 27.0 (iPhone 18 Pro sim) | 50.3 / 50.8 / 51.8 s | 1.3 s |
| iOS 26.2 (iPhone 17 Pro sim) | 0.48 s | 0.46 s |

`ObjectSubclassingControllerTests`, `ObjectSubclassTests`, `ObjectSubclassPropertiesTests`: 34/34 pass on the iOS 26.2 simulator with Xcode 27.0. (The `ParseUnitTests-iOS-host` test host cannot currently launch on the iOS 27 simulator: it traps in `_UIApplicationEvaluateRuntimeIssueForNoSceneLifecycleAdoption`, unrelated to this change.)

### TODOs before merging

- [x] Existing tests pass
- [ ] Add changes to documentation (none needed; internal behaviour only)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved automatic registration of `PFObject` subclasses.
  - Subclasses are now detected more consistently during startup and when new code is loaded, reducing cases where eligible subclasses were missed.


chrisfoulds:
Hit this too (50s+ stall in [PFUser currentUser] on iOS 27) and shipped a narrower workaround while waiting on a merged fix, in case it's a useful data point for the review here.

Instead of rewriting the scan to use objc_copyClassList(), I kept the existing per-bundle objc_copyClassNamesForImage approach but added an early-exit in _registerSubclassesInLoadedBundle: that skips any bundle whose resolved path isn't under the main app bundle's path, before the expensive per-image call ever runs:

- (void)_registerSubclassesInLoadedBundle:(NSBundle *)bundle {
    if (![bundle isKindOfClass:NSBundle.class]) {
        return;
    }
    // Subclasses can only live in the app or its embedded frameworks, so skip
    // every other bundle before touching it. On iOS 27 each
    // objc_copyClassNamesForImage call walks all ~3000 loaded images, and even
    // -executablePath is costly per bundle - together they stall Parse's first
    // use for up to a minute while it holds its core manager lock.
    // Resolve both sides, as the device can report /var/... for one and
    // /private/var/... for the other - Parse's own PFUser etc. are found by
    // this scan, so Parse.framework must never be skipped.
    NSString *appPath = [[NSBundle mainBundle].bundlePath stringByResolvingSymlinksInPath];
    if (![[bundle.bundlePath stringByResolvingSymlinksInPath] hasPrefix:appPath]) {
        return;
    }
    if (!bundle.loaded || !bundle.executablePath) {
        return;
    }
    if ([bundle.bundlePath hasPrefix:@"/System/"] || [bundle.bundlePath hasPrefix:@"/Library/"] || [bundle.bundlePath rangeOfString:@"iPhoneSimulator.sdk"].location != NSNotFound) {
        return;
    }
    [self _registerSubclassesInBundle:bundle];
}

The O(n²) blowup on iOS 27 comes from calling objc_copyClassNamesForImage (and resolving -executablePath) once per loaded image across ~3000 system frameworks. PFObject subclasses realistically only live in the app's own bundle or its embedded frameworks, so skipping everything outside that path avoids almost all of those calls without touching the actual scanning/registration logic.

Got the stall down from 50-60s to effectively instant on iOS 27, no measurable regression on earlier versions. Narrower than the objc_copyClassList() rewrite here — doesn't help apps loading PFObject subclasses from bundles outside the main app path — but a much smaller, lower-risk stopgap if anyone wants one while the KVO-generated-class concern in this PR gets sorted.

mman:
@chrisfoulds @conanjb Why am I not seeing this in my app, deployed in prod with Xcode 26, and running in iOS 27 simulator with Xcode 27. I do not want say your bug report or fix is invalid or anything, but what's special about your apps that it triggers that and how can I reproduce it?

mtrezza:
Same here; not able to reproduce; could you post a minimal example to reliably reproduce the issue?

chrisfoulds:
This is enough to hang my app for 50-55 secs without fixes.
    
    if([PFUser currentUser])
    {
        [PFUser logOut];
    }
    
    
    if (![PFUser currentUser])
    {
    
    }

